PR #26, opened 2018-04-09
Combo loop handling improvements

chick:
Uses new firrtl digraph ability to capture node
found in combinational loop detection
Better feedback for user


src/test/scala/treadle/GetIntBreakdownSpec.scala
```diff
@@ -53,7 +53,7 @@ class GetIntBreakdownSpec extends FreeSpec with Matchers {
     }
     true
   }
-  "GetIntBreakdownSpec should pass a basic test" in {
+  "GetIntBreakdownSpec should pass a basic test" ignore {
```
ucbjrl: Is this test intentionally ignored (as opposed to a debugging change that was unintentionally committed)?

PR #28, opened 2018-04-25
Some general fixes

chick:
* AssignInt was not calling plugin hook
* Change default clock parameters to be more like verilog
* Necessary improvements to expression rendering
  * invert order so last thing you see is signal being rendered
  * fix so it does not follow the mux branch not taken
  * added tests
* VCD fixes
  * memoize ignored wires
  * make better elimination of temp signal, keep _GEN_ signals for now

### DIFF
--- a/src/main/scala/treadle/executable/ClockInfo.scala
+++ b/src/main/scala/treadle/executable/ClockInfo.scala
@@ -10,6 +10,6 @@ case class ClockInfo(
 
 object ClockInfo {
   val DefaultName: String = "clock"
-  val DefaultPeriod: Long = 1000L
-  val DefaultOffset: Long = 500L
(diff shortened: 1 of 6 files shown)


src/main/scala/treadle/executable/ClockInfo.scala
```diff
@@ -10,6 +10,6 @@ case class ClockInfo(
 
 object ClockInfo {
   val DefaultName: String = "clock"
-  val DefaultPeriod: Long = 1000L
-  val DefaultOffset: Long = 500L
+  val DefaultPeriod: Long = 10L
```
ucbjrl: Could we have a comment (or better yet, some ScalaDoc) on what the units are, or what reasonable values would be?

PR #29, opened 2018-04-30
Migrate to "Annotations as Options"

seldridge:
@chick, @ucbjrl, @jackkoenig: This aligns Treadle with the "Annotations as Options" proposal I've been working on (in the same way that https://github.com/freechipsproject/chisel3/pull/804 aligns Chisel with it).

Fundamentally, this does two things and follows them to their logical conclusion: 
  1. Options are parsed using [scopt's "immutable" approach](https://github.com/scopt/scopt#immutable-parsing) into a `lazy val options: AnnotationSeq` as opposed to [scopt's "mutable" strategy](https://github.com/scopt/scopt#mutable-parsing) (e.g., to directly mutate `var treadleOptions`)
  2. `var treadleOptions` is replaced with `lazy val treadleOptions` determined by examining `options: AnnotationSeq`

This effectively defines a two-stage parsing scheme that can accept mix-ins that define new options. First, an `ExecutionOptionsManager` (with mix-ins defining options) determines an `AnnotationSeq` representing options. Then, each lazy val corresponding to a tool's options (e.g., `treadleOptions`, `firrtlOptions`, `chiselOptions`) will be built by examining the available annotations as needed. Option checking is intended to happen when building these lazy vals, as opposed to by scopt.

To achieve this for Treadle, the following actions are taken:
  - Every option (a member of `TreadleOptions`) must be determined by examining a Treadle-specific annotation (`TreadleOption`)
  - Every annotation can be set with a Treadle command line option
  - Old-style APIs that passed an `OptionsManager` and some other information (like FIRRTL source represented as a string) are either removed or deprecated.

See:
  - FIRRTL issue: https://github.com/freechipsproject/firrtl/issues/764
  - FIRRTL PR: https://github.com/freechipsproject/firrtl/pull/765
  - Chisel PR: https://github.com/freechipsproject/chisel3/pull/804

### DIFF
--- a/src/main/scala/treadle/Driver.scala
+++ b/src/main/scala/treadle/Driver.scala
@@ -2,31 +2,49 @@
 
 package treadle
 
-import firrtl.{ExecutionOptionsManager, HasFirrtlOptions}
+import firrtl.{ExecutionOptionsManager, HasFirrtlExecutionOptions, FirrtlSourceAnnotation}
+import firrtl.annotations.{Annotation, NoTargetAnnotation, SingleStringAnnotation}
 import treadle.executable.ClockInfo
+
+sealed trait TreadleOption { self: Annotation => }
+case object WriteVcdAnnotation extends NoTargetAnnotation with TreadleOption
+case object VcdShowUnderscoredAnnotation extends NoTargetAnnotation with TreadleOption
+case object VerboseAnnotation extends NoTargetAnnotation with TreadleOption
+case object SetOrderedExecAnnotation extends NoTargetAnnotation with TreadleOption
+case object AllowCyclesAnnotation extends NoTargetAnnotation with TreadleOption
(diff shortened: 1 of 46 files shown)


chick:
Looks sensible to me.

seldridge:
Just note: this is breaking the REPL. I'm taking a look at that now... Fundamentally, the problem is you can't interact with a `HasFirrtlExecutionOptions` before it has mandatory annotations/options, e.g., an instance of this needs to know what the FIRRTL source is. The REPL, initially, doesn't have a source before you `load` one in. It's not insurmountable, but I just want to document what was going on here. 

There's some more info here: https://github.com/freechipsproject/firrtl/pull/765#issuecomment-385723829.

seldridge:
Closing this makes sense. Going forward, we should be able to `Stage`-ify this in a separate PR.

PR #42, opened 2018-05-23
Fix assert() macro failure.

ucbjrl:
PR #41 introduced the following compile-time error:
```bash
[info] Compiling 59 Scala sources to .../treadle/target/scala-2.11/test-classes ...
[error] .../treadle/src/test/scala/treadle/TestUtilsSpec.scala:36:29: macro applications do not support named and/or default arguments
[error]       if (count > 18) assert(condition = false)
[error]                             ^
[error] .../treadle/src/test/scala/treadle/TestUtilsSpec.scala:106:29: macro applications do not support named and/or default arguments
[error]       if (count > 18) assert(condition = false)
[error]                             ^
[error] two errors found
[error] (Test / compileIncremental) Compilation failed
```


ucbjrl:
The failing code was (re-?)introduced with PR #41.

PR #50, opened 2018-07-13
Bump to Scala 2.12.6 and make it the default.

ucbjrl:


ucbjrl:
We may want to hold off on this (I should have said something with the
initial PR), until all the BIG5 are changed. It's purely a matter of
convenience.

PR #52, opened 2018-07-17
Makes options and BlackBox support distinct from Interpreter

chick:
Prior to this a project could not depend on both

For options
- all options preceded by "treadle-"
- all abbreviated options preceded by "t"

For black boxes
- cycle => clockChange
- fullName => completeName

Fixed up problem in Memory.scala where memory pipeline registers did not
have associated clocks defined, which could cause exception in verbose
mode.

### DIFF
--- a/src/main/scala/treadle/Driver.scala
+++ b/src/main/scala/treadle/Driver.scala
@@ -41,45 +41,45 @@ trait HasTreadleOptions {
 
   var treadleOptions = TreadleOptions()
 
-  parser.note("firrtl-engine-options")
+  parser.note("treadle-options")
 
-  parser.opt[Unit]("fint-write-vcd")
-    .abbr("fiwv")
+  parser.opt[Unit]("tr-write-vcd")
+    .abbr("tiwv")
     .foreach { _ =>
       treadleOptions = treadleOptions.copy(writeVCD = true)
     }
     .text("writes vcd execution log, filename will be base on top")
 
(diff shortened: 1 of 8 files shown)


ucbjrl:
Could/Should this share any common options with `firrtl-interpreter` (maybe it does)?

ucbjrl:
Right, but it would be nice if the text of identical options could be shared. If I want a vcd file, it would be a shame to have to say `--fint-gen-vcd` if the backend is the `firrtl-interpreter` and `--tr-gen-vcd` if the backend is `treadle`.

chick:
On that note, currently with verilator the default is to write a vcd and instead one must use yet another option to defeat that. :-(

chick:
I'll bring this up with @seldridge, we need to build in some support for this notion

seldridge:
There may be some motivation in providing expected common options that many different backends will want do define. The `--vcd` example is a good candidate for that. What the backends _should_ care about is what annotation they are looking for, e.g., does a `GenerateVcdAnnotation` exist?

However, https://github.com/freechipsproject/chisel3/issues/860 does introduce (at least in my view) the notion of a single toolflow run wanting to run multiple backends. There's then a question of how to disambiguate `--vcd` for the different backends. That may lead us back to `--tr-vcd` and `--fint-vcd`, etc. Alternatively, some approach of disambiguating regionally might work, e.g., `+treadle --vcd +firrtl-interpreter --vcd`, and use annotations that contain information about what `Class[Backend]` they are enabled for.

PR #53, opened 2018-07-20
Support for loading memories from files at startup

chick:
- Provide an update method for loading memory at offset
- Create memory initializer
  - keeps track of memories tagged for loading at start-up
  - initializes them, could be re-done by REPL
- symbol table builds list of modules with memories to load
- bit of additional logic to match memories with modules
- add simple test to validate some of this logic

### DIFF
--- a/src/main/scala/treadle/executable/Memory.scala
+++ b/src/main/scala/treadle/executable/Memory.scala
@@ -6,6 +6,7 @@ import treadle._
 import firrtl.{MemKind, WireKind}
 import firrtl.ir.{ClockType, DefMemory, Info, IntWidth}
 import RenderHelper.ExpressionHelper
+import firrtl.annotations.{ComponentName, LoadMemoryAnnotation, MemoryLoadFileType, ModuleName}
 
 import scala.collection.mutable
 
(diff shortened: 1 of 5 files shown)


src/main/scala/treadle/executable/Memory.scala
```diff
@@ -590,6 +591,52 @@ object Memory {
...
+      case _ => Seq()
+    }
+  }
 
+  private def doInitialize(memorySymbol: Symbol, fileName: String, radix: Int): Unit = {
+    io.Source.fromFile(fileName).getLines().zipWithIndex.foreach {
+      case (line, lineNumber) if lineNumber < memorySymbol.slots =>
+        try {
+          val value = BigInt(line.trim, radix)
+          engine.dataStore.update(memorySymbol, lineNumber, value)
```
ucbjrl: Could we encode the radix with the data representation itself? This seems to me to be more a property of the string representing the value than a property of the memory initializer. I would think to convert between base 10, 16, or 2 representations would just require changing the source file containing those representations. Or maybe I misunderstand `anno.hexOrBinary`.

chick: There are only two functions in verilog that do the file loading to memory. One uses hex-text and the other uses binary-text.  I think this is a familiar choice they are making and it shouldn't be encoded in the file name

chick: @ucbjrl ping

PR #60, opened 2018-10-05
Add some API methods needed for Chisel GUI development

chick:
The chisel gui needs additional access to treadle internals to display waveforms
- Adds several methods to return data (WaveForms) from rollback buffers to the GUI
- windows chokes on history file
  - move default location to the local directory
    - this is small change to behavior
    - but is an improvement because separate projects should have separate history files
  - catch creation errors


src/main/scala/treadle/executable/RenderedExpression.scala
```diff
@@ -135,9 +135,9 @@ class ExpressionViewRenderer(
               symbolTable.inputPortsNames.contains(symbol.name) ||
               symbolsSeen.contains(symbol)
             )) {
-            if(showValues) {
-              symbolsToDo.enqueue(SymbolAtDepth(symbol, displayDepth + 1, dataTime, dataArrays))
-            }
+//            if(showValues) {
```
ucbjrl: Remove commented out code?
```suggestion
//            if(showValues) {
 ...
 //           }
```

chick: comments removed

src/main/scala/treadle/executable/RollBackBuffer.scala
```diff
@@ -30,13 +30,16 @@ class RollBackBufferRing(dataStore: DataStore) {
 
   def newestToOldestBuffers: Seq[RollBackBuffer] = {
     var list = List.empty[RollBackBuffer]
-    var index = latestBufferIndex
-    while(index != oldestBufferIndex) {
-      list = list :+ ringBuffer(index)
-      index -= 1
-      if(index < 0) {
-        index = numberOfBuffers - 1
+    if (latestBufferIndex != oldestBufferIndex) {
```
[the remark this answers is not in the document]
chick: Added many more comments to explain the above is a ring buffer and the != is the appropriate test.

PR #64, opened 2018-11-20
Make sbt a little better in handling terminal interrupt <cntrl-c>

chick:
Still less than ideal.


ucbjrl:
Should we add this to all `build.sbt` files?

jackkoenig:
@ucbjrl we probably should, 'tis a nice feature

PR #65, opened 2018-11-26
Better clock handling

chick:
- Add advanceTime method to API for precise control of multi-clock DUT
- If signal poked a second time without time or clock advance, evaluate circuit
- Fix scalastyle in engine
- Check transition on triggered assign, and update if up transition
- Add tests for use of advanceTime API

### DIFF
--- a/src/main/scala/treadle/TreadleTester.scala
+++ b/src/main/scala/treadle/TreadleTester.scala
@@ -80,6 +80,12 @@ class TreadleTester(input: String, optionsManager: HasTreadleSuite = new Treadle
       new MultiClockStepper(engine = this.engine, clockInfoList, wallTime)
   }
 
+  /**
+    * Advance time in ticks of the [[UTC]] wallTime, the default is picoseconds, but can be
+    * read by the scaleName of the wallTime.  One should probably be advancing by some simple factor
+    * of a clock period. The clockInfoList of the options should define this (could be more than one).
+    * @param interval units are in units of the [[wallTime]] scale.
+    */
   def advanceTime(interval: Long): Unit = {
     assert(interval >= 0L, "TreadleTester#advanceTime called with negative value")
     wallTime.setTime(wallTime.currentTime + interval)
--- a/src/main/scala/treadle/TreadleTester.scala
+++ b/src/main/scala/treadle/TreadleTester.scala
@@ -80,6 +80,11 @@ class TreadleTester(input: String, optionsManager: HasTreadleSuite = new Treadle
       new MultiClockStepper(engine = this.engine, clockInfoList, wallTime)
   }
 
+  def advanceTime(interval: Long): Unit = {
+    assert(interval >= 0L, "TreadleTester#advanceTime called with negative value")
+    wallTime.setTime(wallTime.currentTime + interval)
+  }
+
   /*
   The Idea here is that combinational delay will be used when a peek follows a poke without a step
   This should allow VCD output to show the events as if they had taken place in a small
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -10,6 +10,8 @@ import treadle.chronometry.{Timer, UTC}
 import treadle.utils.{Render, ToLoFirrtl}
 import treadle.vcd.VCD
 
+import scala.collection.mutable
+
 //scalastyle:off magic.number number.of.methods
 class ExecutionEngine(
     val ast             : Circuit,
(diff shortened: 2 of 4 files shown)


src/test/scala/treadle/ClockedManuallySpec.scala
```diff
@@ -0,0 +1,150 @@
...
+    tester.poke("in1", 1)
+    tester.step()
+    tester.poke("in1", 2)
+    tester.step()
+    tester.poke("in1", 1)
+
+    tester.report()
+  }
+
+}
```
ucbjrl: How about some failure cases as well? "updating circuit on second update of same input without clock advance"

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -215,10 +228,20 @@ class ExecutionEngine(
...
         vcd.wireChanged(symbol.name, dataStore(symbol), symbol.bitWidth)
       }
+
+      if(
+        scheduler.triggeredAssigns.contains(symbol) &&
+        currentValue == Big0 && adjustedValue == Big1
+      ) {
+
+        scheduler.executeCombinationalAssigns()
+        scheduler.executeTriggeredAssigns(symbol)
```
ucbjrl: Is this necessary? If `scheduler.executeCombinationalAssigns()` doesn't clear `executeTriggeredAssigns`, doesn't the previous logic guarantee that `scheduler.triggeredAssigns.contains(symbol)`?
Perhaps I need to look more closely at `executeTriggeredAssigns` and the conditions under which it is cleared.

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -215,10 +228,20 @@ class ExecutionEngine(
...
       dataStore.update(symbol, adjustedValue)
       vcdOption.foreach { vcd =>
         vcd.wireChanged(symbol.name, dataStore(symbol), symbol.bitWidth)
       }
+
+      if(
+        scheduler.triggeredAssigns.contains(symbol) &&
+        currentValue == Big0 && adjustedValue == Big1
+      ) {
+
```
ucbjrl: This detects the rising edge of a poked signal? Maybe add a comment to that effect. Are `Big0` and `Big1` the only legal values, or is this implying the signal is some kind of clock?

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -200,6 +204,15 @@ class ExecutionEngine(
     val symbol = symbolTable(name)
 
     inputsChanged = true
+    if(symbolsPokedSinceEvaluation.contains(symbol)) {
+      if(verbose) {
+        println(s"updating circuit on second update of same input without clock advance")
```
ucbjrl: The default is to silently accept multiple updates, correct?

src/main/scala/treadle/TreadleTester.scala
```diff
@@ -80,6 +80,11 @@ class TreadleTester(input: String, optionsManager: HasTreadleSuite = new Treadle
       new MultiClockStepper(engine = this.engine, clockInfoList, wallTime)
   }
 
+  def advanceTime(interval: Long): Unit = {
```
ucbjrl: How about some ScalaDoc - what units is `interval` measured in?

src/main/scala/treadle/executable/ExecutionEngine.scala
chick: triggeredAssigns are not cleared, they are simply a list of clock or clock like things where a particular transition could change registers or other downstream logic.  executeCombinational goes first just to make sure that any other inputs or things hanging around are up to date to before the triggeredAssigns code, e.g. that the register input is there before the clock latches that value into the register output. It is possible that there is some over freshening of the circuit, but the current state of treadle is a fairly complex  balance between trying not to run unnecessary evaluation and not missing anything any calculation that happens because of derived clocks etc.

src/main/scala/treadle/executable/ExecutionEngine.scala
chick: Yes, the trigger symbol (the key) of a triggeredAssign must be a clock-like (bool) signal.

src/main/scala/treadle/executable/ExecutionEngine.scala
chick: Yes, this is an edge case that appears in testers-2 where richard is not calling step but instead just toggling an ordinary bool that is cast to a clock inside a module. I'm not that happy with the idiom, but it makes some sense, and we are bit straight-jacketed by the existing backend APIs

PR #66, opened 2018-11-29
Async reset 1

chick:
Support for black boxes that provide a sync reset
- AsyncResetReg example provided, derived from rocket-chip
- Comments and todo added on BlackBoxCycler
- New ExternalModuleInputAssigner informs black box of input changes
  - wraps assigner that updated treadle's wire going into black box
- Keep tracks of black box inputs found during compilation
- addAssigner() adds ExternalModuleInputAssigner wrapper where needed
- Fix broken indentation in addAssigner
- Fix setParams being called in two places, not just happens in compile
- Minor fix of unused match variables
- Extends ScalaBlackBox with inputChanged
- Made test_run_dir default directory when manager unspecified.
  - Cleans up many messy files in project root
- New tests demonstrates Asynchronous reset capability of black box


src/main/scala/treadle/executable/BlackBoxCycler.scala
```diff
@@ -5,6 +5,16 @@ package treadle.executable
...
+/**
+  * Implements an assigner that can be scheduled to publish clock transitions
+  * to specific black box implementations
+  * @param symbol symbol name of instance
+  * @param blackBox the instance
+  * @param clockSymbol clock used by instance
+  * @param dataStore the data
+  * @param info source location
+  */
+//TODO: This should be called everytime something hanppens to clock and should indicate all possible transitions
```
ucbjrl: Is it the `run` method that should be called? It would be useful to have an example of how to use this from client code.

src/main/scala/treadle/executable/BlackBoxCycler.scala
```diff
@@ -5,6 +5,16 @@ package treadle.executable
 import firrtl.ir.Info
 import treadle.ScalaBlackBox
 
+/**
+  * Implements an assigner that can be scheduled to publish clock transitions
+  * to specific black box implementations
+  * @param symbol symbol name of instance
```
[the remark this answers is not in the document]
chick: The instance is a pointer to a scala object which is the implementation of the black box, i.e. the specific code that is called when a black box's inputs are changed or output are used. That black box has a symbol in order to reference to it, the  symbol is there because it is required for all Assigners largely for debugging purposes.

src/main/scala/treadle/executable/BlackBoxCycler.scala
chick: All `Assigner`s have a run method in order to update the wire they are associated with, BlackBox clocks use this *magic* assigner, who's run method is only called when the associated clock changes. `run` is an internal API and not documented itself, but by convention is documented (usually :-) in the documentation of the Assigner class, as it is here.

PR #82, opened 2019-01-24
Remove references to firrtl.PrimOps.Shl - goodbye Shlw

ucbjrl:
With freechipsproject/firrtl#994, `Shlw ` node is no more. Long live `Shlw `


aswaterman:
I got confused by the title of this PR because we only got rid of the `Shlw` working-IR node, not `Shl`, but the implementation looks right.

Issue #85, opened 2019-02-12
Treadle gives poor performance with large memory


sashimi-yzh:
I use [this version of chisel-template](https://github.com/freechipsproject/chisel-template/tree/014630fbcf5a8caf082a9c0fd699d7486f4d2429) to develop a single-cycle CPU. I use the following statement to create a memory which the CPU is going to access.
```
val mem = Mem(memByte, UInt(8.W))
```
I run chisel-tester with the `treadle` backend. The tester is something like
```
do {
  step(1)
  state = peek(top.io.state)
} while (state != END)
```

**But I observe that when `memByte` grows, `treadle` gives poor performance.**

I start by running my CPU with a small program. This program require 64KB memory to run. And it ends after about `120000` cycles. Then I double the value of `memByte`, and run the program again. Note that doubling the memory size does not affect the program behavior. I collect the real world time cost during simulation, and the simulation frequency. This data is reported by `treadle` after the simulation ends.

```
===== 120000 cycles  (-J-Xmx512M -J-Xms512M) =====

64KB -> 9.294537 seconds 12848.41 Hz
128KB -> 12.573144 seconds 9506.77 Hz
256KB -> 23.733160 seconds 5043.37 Hz
512KB -> 61.044713 seconds 1962.58 Hz
1MB -> 144.135788 seconds 833.48 Hz
```
This shows that when the size of memory doubles, the real world time also doubles, meanwhile the frequency drops to half.

When the memory size grows to 1MB, it takes a long time to finish the program. Therefore I modify the tester to let the simulation end after `12000` cycles:
```
do {
  step(1)
  state = peek(top.io.state)
  nrInstr +=1
} while (state != END && nrInstr < 12000)
```
Then I continue to collect data.
```
===== 12000 instructions (-J-Xmx512M -J-Xms512M) =====

1MB -> 15.067397 seconds 796.75 Hz
2MB -> 29.217526 seconds 410.88 Hz
4MB -> 57.313398 seconds 209.46 Hz
8MB -> 118.053021 seconds 101.69 Hz
```

Again, when the memory size grows to 8MB, it takes about 2 minutes to finish the simulation. I further reduce the number of simulation cycles.
```
===== 1200 instructions (-J-Xmx512M -J-Xms512M) =====

8MB -> 12.139967 seconds 99.26 Hz
16MB -> "java.lang.OutOfMemoryError: Java heap space"
```
But when I set the memory size to 16MB, I get the following exception.
```
[...]

PR #87, opened 2019-02-13
Set default rollbackBuffers to 0 to improve memory performance

chick:
- Changed default value to 0
- Added another feature to Regression that demonstrates above change
- Added small test to answer question about sign extension of SInt combined with And

### DIFF
--- a/src/test/scala/treadle/primops/AndOrXor.scala
+++ b/src/test/scala/treadle/primops/AndOrXor.scala
@@ -188,4 +188,25 @@ class AndOrXor extends FreeSpec with Matchers {
       AndInts(() => a, () => b, bitWidth).apply() should be(expected)
     }
   }
+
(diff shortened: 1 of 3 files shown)


chick:
This fixes a problem with performance reported in Issue #85

src/test/scala/treadle/primops/AndOrXor.scala
```diff
@@ -188,4 +188,25 @@ class AndOrXor extends FreeSpec with Matchers {
       AndInts(() => a, () => b, bitWidth).apply() should be(expected)
     }
   }
+
+  "(7.U.asSInt & true.B.asSInt).asUInt => 7.U" in {
```
ucbjrl: Should this be a separate PR?

chick: yes, removed it!

PR #88, opened 2019-02-25
Fix for problem with black boxes with state interconnected

chick:
- Extend ScalaBlackBox API to add ability to report more dependencies
- SymbolTable uses new method to build better dependency graph
- Added tests for mutually connected registers
- Added similar tests for mutually connected ASyncReset black box registers

### DIFF
--- a/src/test/scala/treadle/RegisterSpec.scala
+++ b/src/test/scala/treadle/RegisterSpec.scala
@@ -345,4 +345,67 @@ class RegisterSpec extends FlatSpec with Matchers {
     tester.expect("reg2", 54)
 
   }
+
+  behavior of "mutually connected registers"
+
+  it should "alternate values" in {
+    val input =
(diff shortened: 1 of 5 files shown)


src/test/scala/treadle/RegisterSpec.scala
```diff
@@ -345,4 +345,67 @@ class RegisterSpec extends FlatSpec with Matchers {
...
+      """.stripMargin
+
+    val manager = new TreadleOptionsManager {
+      treadleOptions = treadleOptions.copy(
+        showFirrtlAtLoad = true,
+        setVerbose = false,
+        lowCompileAtLoad = true
+      )
+      firrtlOptions = firrtlOptions.copy(
+        noDCE = true,
```
ucbjrl: Trailing comma (Scala 2.11)

Issue #90, opened 2019-03-28
Incorrect Waveform Output


nullromo:
I'm seeing a non-deterministic value in my waveform after peeking and poking repeatedly without stepping. However, the print output is correct.

Here is some stripped-down code that demonstrates the behavior:

```scala
package example

import chisel3.{Driver => _, _}
import chisel3.iotesters._
import chisel3.util.log2Ceil

/**
  * Tester for memory tester.
  */
class MemoryTester(c: MemoryTestModule) extends PeekPokeTester(c) {
  // create some dummy data
  val writeData: Seq[Int] = Seq(123, 1, 5, 18, 99, 23, 7, 6)

  // put the data into the memory
  for ((d, a) <- writeData.zipWithIndex)
    write(a, d)

  // read back all the data
  for ((_, a) <- writeData.zipWithIndex)            // COMMENT A
    read(a)

  // read a write simultaneously
  poke(c.io.writeAddress, 0)
  poke(c.io.writeData(0), 700)
  poke(c.io.writeEnable(0), true)
  poke(c.io.readAddress, 1)                 // COMMENT B
  println("Before step " + peek(c.io.readAddress) + ":" + peek(c.io.readData))
  step(1)
  println("After step " + peek(c.io.readAddress) + ":" + peek(c.io.readData))

  /**
    * Writes some data at some address.
    */
  def write(address: Int, data: Int): Unit = {
    poke(c.io.writeAddress, address)
    poke(c.io.writeData(0), data)
    poke(c.io.writeEnable(0), true)
    println("Writing " + address + ":" + data)
    step(1)
  }

  /**
    * Reads the data at some address.
    */
  def read(address: Int): Unit = {
    poke(c.io.readAddress, address)
    if (c.p.syncRead)
      step(1)
    val data = peek(c.io.readData)
    println("Reading " + address + ":" + data)
    expect(data == writeData(address), "Data out was not data in.")
  }

}

/**
  * Spec for memory tester.
  */
class MemoryTesterSpec extends ChiselFlatSpec {
  val dutGen: () => MemoryTestModule = () => MemoryTestModule(MemoryTestParameters(256, 48, syncRead = false))
  "MemoryTest" should "work properly" in {
    Driver.execute(TesterArgs() :+ "MemoryTest", dutGen) {
      c => new MemoryTester(c)
    } should be(true)
  }
}

/**
  * Parameters for basic memory module.
  *
  * @param numEntries number of entries in the memory.
[...]

PR #91, opened 2019-03-29
Vcd output broken when poking same wire without clock step. Issue #90

chick:
- Make it so values recorded at given time retains only last value
- Change hashCode and equals of Change class to be based on wire id
- Remove change for given wire before adding new one
- Add a test for last value kept during add behavior
- Fixes Issue #90 Incorrect Waveform Output


chick:
@ucbjrl Doc changes made

PR #99, opened 2019-05-14
More multi clock

chick:
Added test to illustrate use of multiple clocks
and that they work right.
Turns out they don't work if ClockPeriod is not an even number
Add assertion forcing even number to clock creation
Made test for that


chick:
superseded by PR #104

PR #100, opened 2019-05-19
Stage phases 1

chick:
Convert Treadle to use Stage/Phases technology.
Implements replacement for Driver
Converts many tests to new API
Still supports old API


src/main/scala/treadle/VcdReplayTester.scala
```diff
@@ -82,61 +91,90 @@ class VcdReplayTester(optionsManager: VcdReplayTesterOptions) extends LazyLoggin
...
-    val optionsManager = new VcdReplayTesterOptions
+trait VcdReplayTesterCli { this: Shell =>
+  parser.note("VcdReplayTester Front End Options")
 
-    if (optionsManager.parse(args)) {
-      val repl = new VcdReplayTester(optionsManager)
-      repl.run()
-    }
+
+}
```
seldridge: This doesn't appear to provide any options. Should this be removed?

src/main/scala/treadle/TreadleTester.scala
```diff
@@ -433,6 +449,14 @@ object TreadleTester {
     * @return
     */
   def apply(input : String, optionsManager: HasTreadleSuite = getDefaultManager): TreadleTester = {
-    new TreadleTester(input, optionsManager)
+    val sourceAnnotation = FirrtlSourceAnnotation(input)
+    TreadleTester(sourceAnnotation +: optionsManager.toAnnotationSeq)
+  }
+
+  def apply(annotations: AnnotationSeq): TreadleTester = {
```
seldridge: While it does not currently do this, I think that `TransformLike[A]` will, in the future, extend `(A) => A`. Can you use a different method, other than apply here? Alternatively, if you do use an apply, it should be `(AnnotationSeq) => AnnotationSeq`.

src/main/scala/treadle/VcdReplayTester.scala
```diff
@@ -82,61 +91,90 @@ class VcdReplayTester(optionsManager: VcdReplayTesterOptions) extends LazyLoggin
...
+
+class VcdReplayTesterStage extends Stage {
+  val shell = new Shell("vcd_replay_tester") with VcdReplayTesterCli with FirrtlCli
+
+  def apply(annotationSeq: AnnotationSeq): VcdReplayTester = {
```
seldridge: I would avoid an apply that is not `(AnnotationSeq) => AnnotationSeq`.

src/main/scala/treadle/VcdReplayTester.scala
```diff
@@ -82,61 +91,90 @@ class VcdReplayTester(optionsManager: VcdReplayTesterOptions) extends LazyLoggin
...
+  * makes a more convenient way of specifying firrtl source
+  */
+case object VcdReplayFirrtlSource extends NoTargetAnnotation with VcdReplayTesterOptions with HasShellOptions {
+  val options = Seq(
+    new ShellOption[String](
+      longOption = "firrtl-source",
+      shortOption = Some("fs"),
+      toAnnotationSeq = (s: String) => Seq(FirrtlSourceAnnotation(s)),
+      helpText = "firrtl source file to load on startup"
+    )
```
seldridge: If you want, you can use `helpValueName: Option[String]` to provide the same `<firrtl-source-file>`. Do you think it makes sense to include those? Ditto for the ones below.

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -436,77 +443,60 @@ class ExecutionEngine(
...
+    val circuit = annotationSeq.collectFirst{ case TreadleCircuitStateAnnotation(c) => c }.get.circuit
+    val blackBoxFactories = annotationSeq.collectFirst{ case BlackBoxFactoriesAnnotation(bbf) => bbf }.getOrElse(
+      Seq.empty
+    )
+    val allowCycles = annotationSeq.exists { case AllowCyclesAnnotation => true; case _ => false }
+    val rollbackBuffers = annotationSeq.collectFirst{ case RollBackBuffersAnnotation(rbb) => rbb }.getOrElse(
+      TreadleDefaults.RollbackBuffers
+    )
+    val validIfIsRandom  = annotationSeq.exists { case ValidIfIsRandomAnnotation => true; case _ => false }
+    val verbose  = annotationSeq.exists { case VerboseAnnotation => true; case _ => false }
```
seldridge: Performance implications of repeated traversal?

src/main/scala/treadle/stage/TreadleStage.scala
```diff
@@ -0,0 +1,37 @@
+// See LICENSE for license details.
+
+package treadle.stage
+
+import firrtl.AnnotationSeq
+import firrtl.options.{Phase, Shell, Stage}
+import firrtl.stage.FirrtlCli
+import treadle.stage.phases.{CreateTester, GetFirrtlAst, PrepareAst, SetImplicitOutputInfo}
+
+class TreadleCompatibilityStage extends Stage {
```
seldridge: Can this be made `private[treadle]`?

src/main/scala/treadle/stage/phases/GetFirrtlAst.scala
```diff
@@ -0,0 +1,33 @@
...
+        sourcesFound += fs
+        val circuit = Parser.parse(firrtlText)
+        Some(TreadleCircuitAnnotation(circuit))
+
+      case other => Some(other)
+    }
+
+    if(sourcesFound.length > 1) {
+      throw TreadleException(s"Multiple sources of firrtl found for Treadle" + sourcesFound.mkString("\n", "\n", ""))
+    }
```
seldridge: Interesting... I generally put all the checking in one phase. However, this is a perfectly viable alternative and perhaps puts information closer to where it makes sense. (I say leave this as is.)

seldridge: For this check, should this be checking on multiple `TreadleCircuitAnnotation`s additionally or alternatively? What would happen if a user a passes in a `TreadleCircuitAnnotation` and a `FirrtlSourceAnnotation`?

src/main/scala/treadle/stage/phases/PrepareAst.scala
```diff
@@ -0,0 +1,29 @@
...
+    annotationSeq.flatMap {
+      case TreadleCircuitAnnotation(circuit) =>
+        val state = CircuitState(circuit, HighForm, annotationSeq)
+        val newState = transforms.foldLeft(state) { case (prevState, transform) => transform.runTransform(prevState) }
+        Some(TreadleCircuitStateAnnotation(newState))
+      case other =>
+        Some(other)
+    }
+  }
+}
```
seldridge: The way I read this, this has an assumption that there is only one `TreadleCircuitAnnotation`. Is that correct? If so, can you add an `@note` to the Scaladoc indicating as such? If this is intended to work for multiple `TreadleCircuitAnnotation`s then I think this would need to use the FIRRTL compiler grammar idea grouping annotations.

src/main/scala/treadle/stage/phases/SetImplicitOutputInfo.scala
```diff
@@ -0,0 +1,37 @@
...
+  override def transform(annotationSeq: AnnotationSeq): AnnotationSeq = {
+    val name = annotationSeq.collectFirst {
+      case TopNameAnnotation(topName) =>
+        topName
+    }.getOrElse {
+      annotationSeq.collectFirst {
+        case TreadleCircuitAnnotation(circuit) =>
+          circuit.main
+        case TreadleCircuitStateAnnotation(state) =>
+          state.circuit.main
```
seldridge: Should this be allowed or should this be an exception? When `SetImplicitOutputInfo` runs, I think there should only be a `TreadleCircuitAnnotation` and not a `TreadleCircuitStateAnnotation`.

src/test/scala/treadle/DynamicMemorySearch.scala
```diff
@@ -55,130 +56,118 @@ class DynamicMemorySearch extends FreeSpec with Matchers {
...
-
-        // println(s"Done wait count is $waitCount done is ${peek("io_done")} " +
-        //   s"got ${peek("io_target")} got $expectedIndex")
-        expect("io_done", 1)
-        expect("io_target", expectedIndex)
+      var waitCount = 0
+      while(waitCount <= n && tester.peek("io_done") == Big0) {
+        // println(s"Waiting for done $waitCount")
+        // println(this.engine.circuitState.prettyString())
+        // println(s"external list ${list.mkString(",")}")
```
seldridge: Nit: I'd prefer to either keep comments out or to convert them to `info` that prints.

src/test/scala/treadle/DynamicMemorySearch.scala
```diff
@@ -55,130 +56,118 @@ class DynamicMemorySearch extends FreeSpec with Matchers {
...
-        poke("io_en", 1)
-        poke("io_isWr", 1)
-        poke("io_wrAddr", write_address)
-        poke("io_data",   write_address)
-        list(write_address) = write_address
-        step()
-        println(s"Initializing memory address $write_address with $write_address mem($write_address)" +
-                s" is ${peekMemory("list", write_address)}")
+      // println(s"Done wait count is $waitCount done is ${tester.peek("io_done")} " +
+      //   s"got ${tester.peek("io_target")} got $expectedIndex")
```
seldridge: Nit: Comment out or convert to `info`?

src/test/scala/treadle/GetIntBreakdownSpec.scala
```diff
@@ -5,6 +5,14 @@ package treadle
...
 
 
+/**
+  * The author of this fine test, does not quite remember what it is designed to do.
+  * As it currently stands it is a decent demonstration of dead code removal. All but one
+  * of the nodes generated at line 36 are dead-ends and do not contribute to the output.
+  * To see this uncomment the println at line 44 and set showFirrtlAtLoad to true at line 48
+  *
+  * The other point of this test is to verify that the older form of writing a test using anonymous subclasses works
+  */
```
seldridge: A++ Scaladoc. 10/10

src/test/scala/treadle/VecSpec.scala
```diff
@@ -31,20 +33,13 @@ class VecSpec extends FreeSpec with Matchers {
...
-      )
-      commonOptions = commonOptions.copy(
-        targetDirName = "test_run_dir/vec_spec1",
-        topName = "vec_spec"
-      )
-    }
+    val options = Seq(
+      WriteVcdAnnotation,
+      RollBackBuffersAnnotation(20),
+      TopNameAnnotation("vec_spec_1")
```
seldridge: Do you want a `TargetDirAnnotation` here?

src/test/scala/treadle/blackboxes/AsyncResetRegSpec.scala
```diff
@@ -242,19 +238,26 @@ class AsyncResetRegSpec extends FreeSpec with Matchers {
...
+//        showFirrtlAtLoad = false,
+//        writeVCD = true,
+//        setVerbose = true
+//      )
+//      commonOptions = commonOptions.copy(
+//        targetDirName = "test_run_dir/async_reg_to_reg",
+//        topName = "async_reg_to_reg"
+//      )
+//    }
+//    val tester = TreadleTester(input, manager)
```
seldridge: Remove.

src/test/scala/treadle/real/BlackBoxRealSpec.scala
```diff
@@ -80,7 +80,14 @@ class BlackBoxRealSpec extends FreeSpec with Matchers {
         randomSeed = 0L
       )
     }
-    val tester = new TreadleTester(input, optionsManager)
+//    val tester = TreadleTester(input, optionsManager)
```
seldridge: Remove.

src/test/scala/treadle/vcd/VCDSpec.scala
```diff
@@ -151,7 +155,14 @@ class VCDSpec extends FlatSpec with Matchers with BackendCompilationUtilities {
       treadleOptions = treadleOptions.copy(writeVCD = true)
       commonOptions = CommonOptions(targetDirName = "test_run_dir")
     }
-    val engine = new TreadleTester(input, manager)
+//    val engine = TreadleTester(input, manager)
```
seldridge: Remove.

src/test/scala/treadle/vcd/VCDSpec.scala
```diff
@@ -230,15 +241,23 @@ class VCDSpec extends FlatSpec with Matchers with BackendCompilationUtilities {
       treadleOptions = treadleOptions.copy(writeVCD = true, vcdShowUnderscored = hasTempWires)
       commonOptions = CommonOptions(targetDirName = "test_run_dir/vcd_register_delay")
     }
-    {
-      val engine = new TreadleTester(input, manager)
+//      val engine = TreadleTester(input, manager)
```
seldridge: Remove.

src/main/scala/treadle/TreadleOptions.scala
```diff
@@ -0,0 +1,282 @@
...
+sealed trait TreadleOption extends Unserializable { this: Annotation => }
+
+/**
+  * Tells treadle to write a vcd file during simulation
+  */
+case object WriteVcdAnnotation extends NoTargetAnnotation with TreadleOption with HasShellOptions {
+  val options: Seq[ShellOption[_]] = Seq(
+    new ShellOption[Unit](
+      longOption = "tr-write-vcd",
+      shortOption = Some("tiwv"),
```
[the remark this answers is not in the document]
chick: I'm not sure what that alternative is here. I'd love to get rid of the alphabet soup. Are you suggesting a set prefix for each CLI or dropping short forms altogether

README.md
```diff
@@ -32,7 +32,7 @@ To do so, in your project `build.sbt` add a dependency on
 ```
 There are a number of different ways to specify this dependency in the build.sbt file. If you have based your circuit on the 
 [Chisel-template](https://github.com/freechipsproject/chisel-template.git) the addition should look like
-```scala
+```
```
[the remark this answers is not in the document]
chick: No idea why, I'll probably just put it back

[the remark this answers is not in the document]
chick: I'm putting it back, I removed it because IntelliJ was complaining that the scala code in there was broken, but I don't think that's enough reason to change it.

src/main/scala/treadle/TreadleOptions.scala
```diff
@@ -0,0 +1,282 @@
...
+      toAnnotationSeq = _ => Seq(ValidIfIsRandomAnnotation),
+      helpText = "validIf returns random value when condition is false"
+    )
+  )
+}
+
+/**
+  *  Sets the number of rollback buffers in simulator, useful to see why wires have their values
+  */
+//scalastyle:off magic.number
```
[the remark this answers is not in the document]
chick: The variable involved here was originally not zero (zero is not a magic number)
I'll remove the override

src/main/scala/treadle/TreadleOptions.scala
```diff
@@ -0,0 +1,282 @@
...
+import firrtl.ir.Circuit
+import firrtl.options.{HasShellOptions, ShellOption, Unserializable}
+import firrtl.stage.FirrtlSourceAnnotation
+import treadle.executable.{ClockInfo, TreadleException}
+
+sealed trait TreadleOption extends Unserializable { this: Annotation => }
+
+/**
+  * Tells treadle to write a vcd file during simulation
+  */
```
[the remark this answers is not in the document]
chick: I've got over 60 of these, I've always used this convention. I don't mind changing but it seems to me like it would just clutter this PR, I'll make a note to look at this in a later style PR

src/main/scala/treadle/TreadleOptions.scala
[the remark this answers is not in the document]
seldridge: Dropping short forms altogether.

This is still backwards compatible as you have the old driver/execution options if (for whatever reason) you really want to use the old short options. We'd then break backwards compatibility by removing the drivers in the x.3 major release. For a stage/phase port, because backwards compatibility is preserved in the old driver, you have a free hand here. Don't feel like you're backed into a corner with the old short options. As far as I'm concerned, you have a free hand to define these however you want. If you're cool with it as is, then leave it.

I do agree that there should be a way for a stage to define a prefix and all options get that. That would have to start as a FIRRTL PR and percolate out, though.

src/test/scala/treadle/VecSpec.scala
chick: No, I have fixed the internal treadle tests harness to default to test_run_dir, keeps the root cleaner

ucbjrl:
Unfortunately, this has four Scala 2.12-isms and won't compile under Scala 2.11

seldridge:
Ugh... I caught one of these and I still approved. :grimacing: 

This may motivate putting Treadle under CI?

chick:
I have made a fix for two of these that I found under PR #104 
Is that good enough or does it create problems

PR #104, opened 2019-06-04
More multi clock 2

chick:
Minor update for multi-clock
removes older ignored test
uses annotations to control test


chick:
Oops, had a fix, just pushed it now.

chick:
Added fixes to dangling commas

PR #107, opened 2019-07-24
Fix samples that are collisions with

chick:
case based file names are colliding in samples. 
These file are not used except in testing


chick:
This is not fixing the problem

PR #108, opened 2019-07-24
Fix up samples dir

chick:
Samples directory had several files that were duplicates on case insensitive file systems
This deletes or renames those


chick:
@ucbjrl renamed core-simple to reflect contents
got rid of rocket, it is too old

PR #109, opened 2019-07-24
Abstract data store

ekiwi:
These four commits try to decouple the concrete implementation of the `DataStore` from the rest of `treadle`. The eventual goal of my work is to be able to swap out the data type that is used for simulations.

It would be great if you could have a look these first steps, @chick .

### DIFF
--- a/src/main/scala/treadle/executable/DataStore.scala
+++ b/src/main/scala/treadle/executable/DataStore.scala
@@ -605,7 +605,12 @@ object DataStore {
   }
 }
 
-trait HasDataArrays {
+trait DataReader {
+  def apply(symbol: Symbol): Big
+  def apply(symbol: Symbol, offset: Int): Big
+}
+
+trait HasDataArrays extends DataReader {
   def intData  : Array[Int]
   def longData : Array[Long]
   def bigData  : Array[Big]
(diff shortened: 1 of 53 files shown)


src/main/scala/treadle/executable/DataStore.scala
```diff
@@ -621,24 +616,62 @@ object DataStore {
...
+  val plugins : mutable.HashMap[String, DataStorePlugin]   // TODO: change to getter
+  val activePlugins : mutable.ArrayBuffer[DataStorePlugin] // TODO: change to getter
+
+}
+
+trait DataBuffer extends DataReader {
+  def getTime: Long
+}
+
+trait AbstractDataStore extends DataReader with DataWriter with DataSerializer with DataPluginHost {
```
edwardcwang: Random comment - might be good to have some ScalaDoc here.

ekiwi: Do you think it makes sense to split the different functionality expected from our `AbstractDataStore` into different traits, or should all/some of these be unified directly into `AbstractDataStore`?

ekiwi:
Now rebased on #111

ekiwi:
no longer relevant

PR #110, opened 2019-07-25
Abstract rollback buffers

ekiwi:
On top of the abstractions suggested in #109 , this decouples the `RollbackBufferManager` from the concrete `DataStore` used. This is the first commit in this series where I am somewhat skeptical about how it might affect performance.

@chick, is there a reliable way to test `treadle` for performance regressions?


ekiwi:
no longer relevant

PR #111, opened 2019-07-26
Fix deprecations

chick:
Treadle had several hundred deprecation warning
This PR reduces those warnings to < 30. 
There should be now functional changes here.


ekiwi:
@chick, could you also add the `-deprecation` flag to the `build.sbt`?
This way it should be easier to avoid introducing new warnings in the future.

chick:
@ekiwi Good idea. Done. Probably should be done in a number of other projects.

src/main/scala/treadle/executable/PrintfOp.scala
```diff
@@ -74,4 +74,6 @@ object PrintfOp {
   val PrintfOpSymbol = Symbol("printfop", IntSize, UnsignedInt, WireKind, 1, 1, UIntType(IntWidth(1)), NoInfo)
 }
 
-case class PrintInfo(printSymbol: Symbol)
\ No newline at end of file
+case class PrintInfo(printSymbol: Symbol, cardinal: Int) extends Ordered[PrintInfo] {
+  override def compare(that: PrintInfo): Int = that.cardinal - this.cardinal
+}
```
ucbjrl: missing newline at EOF?

PR #113, opened 2019-07-30
Attempt to fix ordering of printf statements

chick:
Make a local dependency between successive printfs on
a per module basis.
Add test to show that this worked

### DIFF
--- a/src/main/scala/treadle/executable/SymbolTable.scala
+++ b/src/main/scala/treadle/executable/SymbolTable.scala
@@ -383,7 +383,7 @@ object SymbolTable extends LazyLogging {
           moduleMemoryToMemorySymbol(moduleMemory) += memorySymbols.head
 
 
-        case stop @ Stop(info, _, clockExpression, enableExpression)   =>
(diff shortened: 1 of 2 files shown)


src/main/scala/treadle/executable/SymbolTable.scala
```diff
@@ -382,7 +383,7 @@ object SymbolTable extends LazyLogging {
           moduleMemoryToMemorySymbol(moduleMemory) += memorySymbols.head
 
 
-        case stop @ Stop(info, args, clockExpression, enableExpression)   =>
+        case stop @ Stop(info, _, clockExpression, enableExpression)   =>
```
ucbjrl: I'm not sure I understand what this has to do with printf

chick: Nothing, it was just an unused instantiation. Makes file perfectly clean.
I can undo if you'd like

ucbjrl: If it's going to be combined with another PR, I would think it would be better to combine it with #114.

PR #119, opened 2019-08-08
Memory: fix en pipeline for read ports

ekiwi:
The `enable` signal on a read port is never used by `treadle` and thus this bug was never discovered. However, I am using it for my symbolic execution extension.

This fixes the fact that instead of creating a enable and a address pipeline the old code would create two address pipelines.
During the course of fixing this I merge the functions for read and write ports into one as their only difference was the latency which can be a function parameter.

### DIFF
--- a/src/main/scala/treadle/executable/Memory.scala
+++ b/src/main/scala/treadle/executable/Memory.scala
@@ -417,33 +417,26 @@ object Memory {
       }
     }
 
-    /*
-      * Makes a read chain of pipeline registers.  These must be ordered reg0/in, reg0, reg1/in ... regN/in, regN
+    /**
+      * Makes a chain of pipeline registers.
+      * These must be ordered reg0/in, reg0, reg1/in ... regN/in, regN
       * This will advance the registers on the specified clock,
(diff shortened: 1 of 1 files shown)


ekiwi:
In case anyone is wondering: the changes from `readerName`/`writerName` to `name` allowed me to keep the 120 character horizontal limit. `name` should not be ambiguous and in the case of a read/write port, it is actually more correct.

src/main/scala/treadle/executable/Memory.scala
```diff
@@ -417,33 +417,25 @@ object Memory {
       }
     }
 
-    /*
+    /**
       * Makes a read chain of pipeline registers.  These must be ordered reg0/in, reg0, reg1/in ... regN/in, regN
```
chick: Maybe this comment should be edited to just saw a chain of pipeline registers

ekiwi: What exactly would you like the comment to say?

chick: "Makes a chain of ..."

PR #123, opened 2019-08-15
Test for async reset

chick:
Add a test that demonstrates treadle's correct behavior with AsyncReset


src/test/scala/treadle/asyncreset/AsyncResetNativeSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    // register returns to input value after reset de-asserted and clock is advanced
+    tester.poke("reset", 0)
+    tester.expect("io_out", 11)
+    tester.step()
+    tester.expect("io_out", 5)
+
+    tester.report()
+  }
+}
```
ucbjrl: missing newline at EOF?

PR #128, opened 2019-08-29
ExpressionCompiler: make manual clocking work with asClock(clk)

ekiwi:
The yosys backend wraps all clock signals in `asClock`.
Not sure why the firrtl type checker allows register clock
signals of type UInt<1> without casting.


ekiwi:
Please do not merge yet. I think there is more to this than I initially thought.

chick:
@ekiwi I have add a DO NOT MERGE label. Just take it off when you'd like me to review.

ekiwi:
I managed to create a failing test case and then add a fix on top.
So from my side this is now ready to merge.

PR #131, opened 2019-09-13
Fix problem with combinational loops

chick:
These were caused by the transform
passes.memlib.VerilogMemDelays in LowFirrtlOptimization
Builds a Treadle specific TreadleLowFirrtlOptimization that
is used instead.


src/main/scala/treadle/stage/phases/PrepareAst.scala
```diff
@@ -17,24 +27,58 @@ trait TreadlePhase extends Phase {
...
+  def outputForm: CircuitForm = LowForm
+
+  def transforms: Seq[Transform] = Seq(
+    passes.RemoveValidIf,
+    new firrtl.transforms.ConstantPropagation,
+    passes.PadWidths,
+    new firrtl.transforms.ConstantPropagation,
+    passes.Legalize,
+    new firrtl.transforms.ConstantPropagation,
+    passes.SplitExpressions,
```
ekiwi: I wonder how getting rid of `SplitExpressions`  would affect performance.
@chick have you every tried if generating fewer intermediate wires improved performance?

chick: I have tried it and it did not seem to make a big difference in execution. It might help more on large circuits, but I don't have that many good examples. riscv-mini, which I am working on to make it use Treadle might provide better ones.

jackkoenig: A DRY-er approach would be to extend the existing `LowFirrtlOptimization` but override the transforms. I took this approach in rfuzz: https://github.com/ekiwi/rfuzz/blob/818ea683c0479245415819eda4baf82c189b65e7/instrumentation/src/main/scala/hardwareafl/firrtltransforms/CustomVerilogCompiler.scala#L117

jackkoenig: You'll notice I have a custom VerilogMemDelays (lol) but it was a different issue where I needed a bugfix before it was available.

PR #134, opened 2019-09-19
Memory: fix memory pipeline types

ekiwi:
This change only becomes visible if you try to run firrtl without `VerilogMemDelays`.

### DIFF
--- a/src/main/scala/treadle/executable/Memory.scala
+++ b/src/main/scala/treadle/executable/Memory.scala
@@ -92,7 +92,7 @@ object Memory {
       sensitivityGraphBuilder.addSensitivity(clk, data)
 
       val pipelineRaddrSymbols = (0 until memory.readLatency).flatMap { n =>
-        buildRegisterTriple(s"$expandedName.$readerString.pipeline_raddr_", n, dataType)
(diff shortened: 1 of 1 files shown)


src/main/scala/treadle/executable/Memory.scala
```diff
@@ -163,12 +163,12 @@ object Memory {
       sensitivityGraphBuilder.addSensitivity(mask, valid)
 
       val pipelineReadDataSymbols = (0 until memory.readLatency).flatMap { n =>
```
albert-magyar: Should the name change to `pipelineReadAddrSymbols` if it's pipelining the address?

ekiwi: Probably yeah. But this code is completely dead anyways since there is no test case that has a combined R/W port. Also: the firrtl passes that treadle currently uses split up all R/W ports and turn them into separate ports.

Addressing the whole issue of R/W ports is outside the scope of this issue, but yeah, this code should be tested+fixed or removed eventually.

ekiwi: The issue that triggered this pull request was the first line I fixed. I only extended it to the other two lines for consistency's sake.

PR #145, opened 2019-10-30
Testers2 boring fix

chick:
- Makes treadle more robust with various forms of firrtl input
- Add Hint that outside programs can identify their firrtl type to treadle
- Add BoreTest to show BoringUtilsSupport
- Add testers creation test to show starting with different forms

### DIFF
--- a/src/main/scala/treadle/TreadleOptions.scala
+++ b/src/main/scala/treadle/TreadleOptions.scala
@@ -2,7 +2,7 @@
 
 package treadle
 
-import firrtl.CircuitState
+import firrtl.{ChirrtlForm, CircuitForm, CircuitState, HighForm, LowForm, UnknownForm}
 import firrtl.annotations.{Annotation, NoTargetAnnotation}
 import firrtl.ir.Circuit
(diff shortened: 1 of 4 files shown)


chick:
@sequencer Can you take a look at this and let me know what you think. This PR I think fixes the problems you have identified and provides tools to make it easier to get the desired behavior from testers2

sequencer:
Yes, I think this should be better than my PR! 
I'm worrying about when firrtl dependency API finally get merged, will `high/middle/low` form become deprecated? Since they won't help on finding out which phases has been executed or not.(just legacy code for transform writing.)
Thus I think(just a shallow idea from my first thought), should we ask user to push final firrtl to treadle(without judging which form it is, assuming it already is low firrtl when feed to treadle), then change the iotester and tester2, making the generation of low firrtl happens there.(to avoid executing 2 round of firrtl compiler).

albert-magyar:
The existing forms will be deprecated for transforms, but they are still
part of the spec.

chick:
Yes, I think testers2 should be changed to call the compiler before launching treadle.
I am preparing a PR to demonstrate that now. 
I should have it ready in the next hour or so

PR #154, opened 2019-11-26
Fixes for problem discovered in riscv-mini

chick:
Shl did not handle cases properly when the result crossed over
into the next larger datatype
- Added logic in oneArgOneParam to manage these cases
- This problem does not apply to other primops in this family
- PadWidths removed from TreadleLowFirrtlOptimization
  - recommendation of Albert and Adam
  - possibly contributing to riscv-mini problems
- Added additional tests to check shift left on data size boundaries

### DIFF
--- a/src/main/scala/treadle/executable/ExpressionCompiler.scala
+++ b/src/main/scala/treadle/executable/ExpressionCompiler.scala
@@ -387,28 +387,51 @@ class ExpressionCompiler(
       val arg1Width = getWidth(expressions.head)
       val param1 = ints.head.toInt
 
-      arg1 match {
-        case e1: IntExpressionResult =>
-          op match {
-            case Head => HeadInts(e1.apply _, takeBits = param1, arg1Width)
-            case Tail => TailInts(e1.apply _, toDrop = param1, arg1Width)
-            case Shl  => ShlInts(e1.apply _, GetIntConstant(param1).apply _)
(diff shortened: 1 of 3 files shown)


src/main/scala/treadle/executable/ExpressionCompiler.scala
```diff
@@ -387,28 +387,51 @@ class ExpressionCompiler(
...
-            case Shr  => ShrBigs(e1.apply _, GetBigConstant(param1).apply _)
-          }
+      if(op == Shl) {
+        (DataSize(getWidth(tpe)), arg1) match {
+          case (IntSize, e1: IntExpressionResult) =>
+            ShlInts(e1.apply _, GetIntConstant(param1).apply _)
+          case (IntSize, e1: LongExpressionResult) =>
+            ShlLongs(e1.apply, GetLongConstant(param1).apply _)
+          case (IntSize, e1: BigExpressionResult) =>
+            ShlBigs(e1.apply _, GetIntConstant(param1).apply _)
```
ucbjrl: Why isn't this:
`ShlBigs(e1.apply _, GetBigConstant(param1).apply _)`
similar to the other `case (_, e1: BigExpressionResult) =>`
?

chick: The shift amount cannot be greater than 2 ^ 31, so I changed them all to be GetIntConstant

PR #155, opened 2019-12-04
Makes treadle printf output formatted more like verilator does

chick:
- creates implicit columnated output by looking at the width of wire
- adds leading zeros to %x fields
- improved performance by parsing format string only once.

### DIFF
--- a/src/main/scala/treadle/executable/PrintfOp.scala
+++ b/src/main/scala/treadle/executable/PrintfOp.scala
@@ -108,62 +108,6 @@ case class PrintfOp(
     (StringContext.treatEscapes(outBuffer.toString()), filters)
   }
 
-  //scalastyle:off method.length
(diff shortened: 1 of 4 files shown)


src/main/scala/treadle/executable/PrintfOp.scala
```diff
@@ -49,29 +72,111 @@ case class PrintfOp(
...
               s = ""
           }
       }
     }
-    StringContext.treatEscapes(outBuffer.toString())
+    (StringContext.treatEscapes(outBuffer.toString()), filters)
   }
+
+  //scalastyle:off method.length
+//  def executeVerilogPrint(formatString: String, allArgs: Seq[Any], bitWidths: Seq[Int]): String = {
```
ucbjrl: perhaps delete the commented-out code?

chick: Good call, done!

PR #157, opened 2019-12-06
The vcd filename can be wrong

chick:
In some situations the vcd filename's root string is wrong
This fixes it to be same as other names created in test_run_dir


chick:
This fixes issue #157

PR #159, opened 2019-12-06
Make printf behavior more closely match verilator

chick:
- Was sometimes off a cycle when using a non-constant enable
- Give printf access to testers wall time to make it easier to match printf to vcd
- Make printf format look a bit more like verilator
- Tangled up with this was some diagnostic code
  - Add an option PrefixPrintfWithWallTime
  - Adds a walltime field to the front of each printf statement
  - This really helps with matching up printfs to VCD views
- There remains a problem with the printf when the reset has gone low at the same time as the clock going high.

### DIFF
--- a/src/main/scala/treadle/TreadleOptions.scala
+++ b/src/main/scala/treadle/TreadleOptions.scala
@@ -206,6 +206,19 @@ case object CallResetAtStartupAnnotation extends NoTargetAnnotation with Treadle
 }
 
 /**
+  *  Tells treadle to prefix printf strings with a wall time
+  */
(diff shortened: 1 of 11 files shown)


chick:
@azidar You and I worked on this before. Change here are that 
- all printf arguments have a register injected in front of them.
- enable also has a register injected in front of it.
   - this produces a problem when reset has transitioned low at the same time as the driving clock goes high.
  - verilator prints in this condition but treadle does not.
  - possibly rework the logic cone of enable to pull reset up to the top and and it with the other conditions
  - as @albert-magyar says, solve the problem generally and analytically

chick:
Just received confirmation from author of Chisel-testers Issue #272 that his PR (using publishLocal) fixed his reported problem

PR #161, opened 2019-12-20
Vcd diff upgrade 2

chick:
Some improvements to vcd diff capability
- Now has human readable dump
- Only considers names it can match between the two files
Fixed up some warning in other vcd stuff
- tailrec annotations
Changed VCD mains to use Stage type launchers

This was also a decent test of merging in changes from a branch that had not been reformatted. That was a bit difficult.


src/main/scala/treadle/vcd/VCD.scala
```diff
@@ -36,15 +36,15 @@ object VCD extends LazyLogging {
...
+  val SectionHeader: Regex = """^\$([^$]+) *$""".r
   val EndSection:    Regex = """^\$end *$""".r
 
   val ScopedModule: Regex = """\s*(?i)(\S+)\s+(\S+)\s*""".r
   val JustScoped:   Regex = """\s*(\S+)\s*""".r
 
   val VarSpec:            Regex = """\s*(\w+)\s+(\d+)\s+(\S+)\s+([\S ]+)\s*""".r
   val ValueChangeScalar:  Regex = """\s*(\d+)(\S+)\s*""".r
-  val ValueChangeVector:  Regex = """\s*([rbh])([0-9\.]+)\s*""".r
+  val ValueChangeVector:  Regex = """\s*([rbh])([0-9.]+)\s*""".r
```
ucbjrl: Is this change really correct? Does `.` lose its special meaning inside character class brackets? It seems logical to do so - otherwise, what would be the difference between `.` and `[.]`?

chick: Yes. dot loses it's wildcard-ness when in brackets.

PR #180, opened 2020-03-05
Updates Treadle to know about new ReadUnderWrite property of DefMem

chick:
- Adds a delay register when memory read port uses ReadUnderWrite.New
- Add test (based on chisel3 test) to show memory follows ReadUnderWrite pattern
- Fix unclosed source in MemoryInitializer


chick:
This has been supplanted by PR #182

PR #182, opened 2020-03-13
Return to using LowFirrtlOptimization

chick:
This will invoke VerilogMemDelays transform that will build
register piplelines, instead of having treadle do it.
This PR changes Memory.scala, to set all latencies to zero
to rely on what was assembled by the VerilogMemDelays.
I'm keeping all the machinery to build these pipelines in Memory.scala since there is a // Todo on VerilogMemDelays in
LowFirrtlOptimization to move it to verilog emitter.

This includes a fix for registers created by Memory to be RegKind instead of WireKind

### DIFF
--- a/src/main/scala/treadle/executable/Memory.scala
+++ b/src/main/scala/treadle/executable/Memory.scala
@@ -17,13 +17,35 @@ limitations under the License.
 package treadle.executable
 
 import treadle._
-import firrtl.{MemKind, WireKind}
-import firrtl.ir.{ClockType, DefMemory, Info, IntWidth}
+import firrtl.{FileUtils, MemKind, RegKind, WireKind}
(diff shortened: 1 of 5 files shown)


ekiwi:
Chick, are you planning to get rid of the delay code in treadle that manually adds the pipeline registers?

chick:
I plan to remove it in the future. Does it affect your work one way or the other?

ekiwi:
> I plan to remove it in the future. Does it affect your work one way or the other?

No. I just suspect that there might still be some bugs in the treadle pipelinging code and its probably best to remove it.

chick:
@ucbjrl I merged your PR into mine. Seems to build ok locally for me. I'm not sure what introduced the problem

PR #194, opened 2020-04-29
Augmented combinational loop report

chick:
When found during symbol table processing It displays a combinational loop
(which will just be the first one found, if there are more than one).
The path contains the statements and the @INFO path with symbols on path in red
If a combinational loop is found by CheckComboLoops and you want to see this enhanced display, turn off the combination loop checking with an annotation or command line flag.
DontCheckCombLoopsAnnotation or --no-check-comb-loops

### DIFF
--- a/src/test/scala/treadle/executable/SymbolTableSpec.scala
+++ b/src/test/scala/treadle/executable/SymbolTableSpec.scala
@@ -21,13 +21,12 @@ import java.io.{ByteArrayOutputStream, PrintStream}
 import firrtl.graph.CyclicException
 import firrtl.stage.FirrtlSourceAnnotation
 import firrtl.transforms.DontCheckCombLoopsAnnotation
-import logger.{LazyLogging, LogLevel, Logger}
 import treadle._
 import org.scalatest.freespec.AnyFreeSpec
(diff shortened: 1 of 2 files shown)


src/test/scala/treadle/executable/SymbolTableSpec.scala
```diff
@@ -16,15 +16,18 @@ limitations under the License.
...
 import firrtl.stage.FirrtlSourceAnnotation
 import firrtl.transforms.DontCheckCombLoopsAnnotation
+import logger.{LazyLogging, LogLevel, Logger}
 import treadle._
 import org.scalatest.freespec.AnyFreeSpec
 import org.scalatest.matchers.should.Matchers
 
 //scalastyle:off magic.number
-class SymbolTableSpec extends AnyFreeSpec with Matchers {
+class SymbolTableSpec extends AnyFreeSpec with Matchers with LazyLogging {
```
ucbjrl, line 30: Should the `LazyLogging` changes be a separate PR?

chick: Done

PR #196, opened 2020-04-30
Provide a way to save lowered Firrtl file

chick:
This adds SaveFirrtlAtLoadAnnotation as a way to see the lowered firrtl file that will be used by Treadle.
Formerly it would just dump file to stdout when ShowFirrtlAtLoadAnnotation was used.
Adds test to show that this works.
Fixed broken command line adding of ShowFirrtlAtLoadAnnotation
One small typo fix in WriteVcdAnnotation

### DIFF
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -16,7 +16,11 @@ limitations under the License.
 
 package treadle.executable
 
+import java.io.PrintWriter
+
 import firrtl.ir.{Circuit, NoInfo}
+import firrtl.options.StageOptions
+import firrtl.options.Viewer.view
 import firrtl.{AnnotationSeq, PortKind}
 import treadle._
 import treadle.chronometry.{Timer, UTC}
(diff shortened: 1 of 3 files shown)


src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -475,8 +475,21 @@ object ExecutionEngine {
   def apply(annotationSeq: AnnotationSeq, wallTime: UTC): ExecutionEngine = {
     val timer = new Timer
     val t0 = System.nanoTime()
+    val stageOptions = view[StageOptions](annotationSeq)
 
     val circuit = annotationSeq.collectFirst { case TreadleCircuitStateAnnotation(c)           => c }.get.circuit
+
```
ucbjrl, line 481: This supports simultaneous output to both a file and stdout. Is this desirable? Is it possible to pass arguments to annotations and use the argument to determine the file name with a name of "" meaning write to stdout?

chick: That makes some sense but I'd prefer to do this under another PR. I think changing it the way you suggest would be a very slight API change or would increase the difficulty of getting the file name right (as it stands it now automatically uses the appropriate tests_run_dir directory and the circuit name).

Issue #185, opened 2020-04-01
Chisel and Treadle need an elegant Plusargs mechanism


davidmlw:
see https://github.com/freechipsproject/chisel3/issues/1396

Rocket-chip use Verilog blackbox plusarg_reader to implement this. While a pure chisel/treadle solution is more elegant, so that Treadle can simulate Rocket-chip firrtl.

chick:
Closing, more specific Issue should be created for any further work on this

PR #197, opened 2020-04-30
First round of support for rocket black boxes

chick:
- ClockDivider2
- ClockDivider3
- EicgWrapper
- PlusArgReader

To access these new Black boxes uses `BlackBoxFactoriesAnnotation(Seq(new BuiltInBlackBoxFactory))`` or
`--tr-add-rocket-black-boxes`
Plusars can be instantiated via PlusArgsAnnotation or `--tr-plus-args`.
Tests are provided for these new black boxes

Also fixed a bug in SymbolTable where black boxes output clocks were not being correctly handled due
to lack of creating a symbol for the previous clock value


chick:
This is the first response to PR #185 
@davidmlw I don't know about elegant but this is a first step towards a treadle support for the array of basic black boxes it uses. I plan on getting to the others reasonably soon, the DTM and Jtag seem like a fair amount of code. Please check it out if you have time. Suggestions welcome

sequencer:
Hi, @chick I wonder if possible to move `ScalaBlackBox` into tester2? since other backends(verilator, vcs, ncsim...) need to share a same codebase of this behavior model.

chick:
I'm not sure if I understand. If you publishLocal this branch you should be able to use it in testers2. The other backends should not be affected

ekiwi:
> since other backends(verilator, vcs, ncsim...) need to share a same codebase of this behavior model.

These other simulators normally use the Verilog Blackbox implementation that comes with RocketChip afaik.

sequencer:
yes, I means if possible to have a unified ScalaBlackBlox for different backends?

chick:
I think that would be a tricky language/environment problem. Conventional black boxes (without Treadle) are verilog sometimes with C/C++ augmentation. Treadle is running in the JVM and thus the ScalaBlackBox was created to be able to emulate the regular black box flow. I think "write black boxes one way" is a hard problem. Do you have some ideas on how to go about this?

sequencer:
Yes, I think will behave like what tester2 currently implemented, Using VPI+RPC connecting to verilator/vcs, and naively connected to treadle.
Currently, internal signal peek/poke/force are not supported in our cpp test harness, but I think it can be done with Annotation renaming during the transform.

ekiwi:
> yes, I means if possible to have a unified ScalaBlackBlox for different backends?

Why would you want to do that? The (or at least some of the) Verilog Black Boxes can be used for synthesis (verifying things before synthesis is what all this simulation is about in the end) whereas the Scala Black Boxes cannot.

sequencer:
> Why would you want to do that.

I think the VIP implementation is a good way to implement a complex model, for example a `TLROM` which can directly read from some file, a `TLRAM` which can introduce arbitrary error.

davidmlw:
> This is the first response to PR #185
> @davidmlw I don't know about elegant but this is a first step towards a treadle support for the array of basic black boxes it uses. I plan on getting to the others reasonably soon, the DTM and Jtag seem like a fair amount of code. Please check it out if you have time. Suggestions welcome

Thanks a lot @chick ! I just saw this, will have a try soon.

chick:
@davidmlw I have some work started on JTAG that has not been pushed. It's a bit farther down in my queue but let me know if you'd like to see it after this get merged.

ekiwi-sifive:
Is this ready to review @chick ?

chick:
Yes, I think it is.

Chick Markley
Staff Programmer -- Adept Lab
University of California, Berkeley
583-2 Soda Hall, Berkeley, CA 94720-1776
chick@berkeley.edu

chick:
Yes, I think it is. I also think @davidmlw has been doing some testing on it as well

src/main/scala/treadle/blackboxes/ClockDividers.scala
```diff
@@ -0,0 +1,126 @@
...
+    Seq.empty
+  }
+
+  /**
+    * Important note: The dependency of io_d on io_q makes this test work by making sure
+    * that the assignments are sorted correctly topologically.
+    * They mirror a similar pattern used on registers, necessary for treadle to be able
+    * to update the circuit in a single pass.
+    * @return
+    */
```
ucbjrl, line 63: misplaced comment? This looks remarkably similar to a comment in AsyncResetBlackBox.scala (which I think is excellent BTW).

src/main/scala/treadle/blackboxes/ClockDividers.scala
```diff
@@ -0,0 +1,126 @@
...
```
ucbjrl, line 120: misplaced comment (as above)?

src/main/scala/treadle/blackboxes/AsyncResetBlackBox.scala
```diff
@@ -0,0 +1,95 @@
...
+package treadle.blackboxes
+
+import firrtl.ir.{Param, Type}
+import treadle._
+import treadle.executable._
+
+/**
+  * Implements a single bit register with asynchronous reset
+  */
+class AsyncResetReg(val instanceName: String) extends ScalaBlackBox {
```
jackkoenig, line 26: Is this necessary anymore? The module in Rocket Chip is no longer a BlackBox: https://github.com/chipsalliance/rocket-chip/blob/2cee947a326459349006ab5aeae8f95e925e3d43/src/main/scala/util/AsyncResetReg.scala

jackkoenig:
I like the idea here, but I'm curious why this code which is Rocket Chip specific is living in treadle itself as opposed to in Rocket Chip? I guess these are general enough-ish that the argument can be made to include them in a Treadle standard library.

src/main/scala/treadle/blackboxes/ClockDividers.scala
chick: Removed comment

src/main/scala/treadle/blackboxes/AsyncResetBlackBox.scala
chick: Removed AsyncResetReg

chick:
I think for now, since I am incrementally doing this and the next steps are harder. I'd rather just keep this local to treadle. If I get it working properly and if it's fast enough to have any utility it might be worth moving to rocket-chip somehow.

ekiwi-sifive:
@chick What is your release strategy for this? Are you internally running tests with RocketChip and we will merge once these tests pass?

chick:
@ekiwi-sifive I don't have one at the moment. There are several more black boxes than need to be built out, and they are a fair amount more complicated than those done so far. I have done some initial work but lately I have been side tracked. But short term I think @sequencer may have some use of the PlusArg black box. Perhaps we should discuss this in chisel dev meeting

davidmlw:
I have a question, how to invode the "+arg=value" from command line? It seems that chisel tools have a convention of "--arg=value", "--arg value", or "-A=value", "-A value".  Can I just use "+arg=value" directly in comand line?

davidmlw:
> I like the idea here, but I'm curious why this code which is Rocket Chip specific is living in treadle itself as opposed to in Rocket Chip? I guess these are general enough-ish that the argument can be made to include them in a Treadle standard library.

I consider treadle should provide a scheme to parse plusargs and pass to internal modules. plusarg_reader is acctually an workaround for chisel to use plusarg of verilog. Maybe firrtlcan provide `$value$plusargs()` as a buildin Op and emit verilog directly, Treadle will just need implement this Op api, and rocket-chip will not need a workaroud of plusarg_reader.

davidmlw (approved):

sequencer:
I think the fundamental thing should be traced back to `DPI`/`VPI` support to simulator.

PR #199, opened 2020-05-06
Fix memory implementation problem with new release

chick:
- Restore treadle's creation of memory latency pipelines when necessary
- Undid forcing of latency values to zero in 3 sections of Memory.scala
- Cleaned up latency tests of memory

I'm a bit perplexed as to how we ended up back here, but everything seems to test properly. non-zero latency values show up in Memory.scala though I thought they would not

### DIFF
--- a/src/main/scala/treadle/executable/Memory.scala
+++ b/src/main/scala/treadle/executable/Memory.scala
@@ -24,19 +24,6 @@ import firrtl.annotations.{ComponentName, LoadMemoryAnnotation, MemoryLoadFileTy
 
 import scala.collection.mutable
 
-////////////////////////////////////////////////////////////////////////////////
-// IMPORTANT NOTE
-// The primary methods here have been effectively disabled by using
-//     val memory = defMemory.copy(readLatency = 0, writeLatency = 0,
-//         readUnderWrite = ReadUnderWrite.Undefined)
-// appearing at the beginning. This is because the functionality contained here,
-// i.e. adding read and write pipeline registers is now being taken care of in the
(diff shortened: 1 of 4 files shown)


albert-magyar:
The write latency of a memory can never be zero. Did you have the example to replicate the issues you were seeing before?

chick:
@ucbjrl produced `MemRegSpec` that errors without these changes. It is based on the Risc example in chisel_tutorial. By forcing latencies to zero in the past, I was just turning off treadle's pipeline creation. But that broke chisel-tutorial

albert-magyar:
> @ucbjrl produced `MemRegSpec` that errors without these changes. It is based on the Risc example in chisel_tutorial. By forcing latencies to zero in the past, I was just turning off treadle's pipeline creation. But that broke chisel-tutorial

I see that, but the issue is that using VerilogMemDelays should cause you to revert to the handling associated with memories with write latency of one cycle, not zero. Based on your code, it looks like you were using pipeline registers even for memories with write latency of one cycle.

I think that in a perfect world, it would be better to avoid adding explicit pipeline registers for memories with write latencies of one cycle. However, given that your original implementation uses pipeline registers for such memories, you will of course see different behavior if you use VerilogMemDelays but then coerce the write latency to zero to remove the pipeline registers. Whether this behavior change is good or bad is beyond my understanding, since I haven't looked too closely at the rest of how Treadle models the behavior of memories.

albert-magyar:
See #200

chick:
Closing. Problem was fixed by PR #200 
Thanks @albert-magyar

ucbjrl:
Closed by #200

PR #215, opened 2020-06-19
Ability to Randomize Registers and Memory

chick:
Seeds are based on register/memory name hashCodes plus a user supplied seed.
This makes randomization deterministic over multiple runs.

Run  RandomizeCircuit and see that happens. Points of interest
Time Event
100  First randomize call
200  Reset
245  Second Randomize call

### DIFF
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -25,7 +25,7 @@ import firrtl.{AnnotationSeq, MemKind, PortKind, RegKind}
 import logger.LazyLogging
 import treadle._
 import treadle.chronometry.{Timer, UTC}
-import treadle.utils.Render
+import treadle.utils.{NameBasedRandomNumberGenerator, Render}
 import treadle.vcd.VCD
 
 import scala.collection.mutable
@@ -109,6 +109,10 @@ class ExecutionEngine(
 
   val timer = new Timer
 
+  if (annotationSeq.contains { RandomizeAtStartupAnnotation }) {
+    randomize()
+  }
+
   if (verbose) {
     if (scheduler.orphanedAssigns.nonEmpty) {
       Render.headerBar(s"Executing static assignments", offset = 8)
@@ -166,17 +170,19 @@ class ExecutionEngine(
     }
   }
 
+  /** Randomize the circuits registers and memories
+    *
+    * @param additonalSeed a seed to move change all the random numbers generated
+    */
   def randomize(additonalSeed: Long = 0L): Unit = {
-    val rand = new Random
+    val randomGenerator = new NameBasedRandomNumberGenerator
 
     val symbolsToDo: Seq[Symbol] = symbolTable.symbols.toSeq
 
     symbolsToDo.foreach { symbol =>
-      rand.setSeed(symbol.name.hashCode + userRandomSeed + additonalSeed)
-      rand.setSeed(rand.nextLong())
 
       def getRandomValue: BigInt = {
-        val big = BigInt(symbol.bitWidth, rand)
+        val big = randomGenerator.nextBigInt(symbol.name, userRandomSeed + additonalSeed, symbol.bitWidth)
         val newValue = if (symbol.dataType == SignedInt) {
           symbol.makeSInt(big, symbol.bitWidth)
         } else {
@@ -184,6 +190,7 @@ class ExecutionEngine(
         }
         newValue
       }
+
       if (symbol.dataKind == RegKind) {
         val newValue = getRandomValue
         logger.info(s"setting ${symbol.name} <= $newValue")
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -21,13 +21,15 @@ import java.io.PrintWriter
 import firrtl.ir.{Circuit, NoInfo}
 import firrtl.options.StageOptions
 import firrtl.options.Viewer.view
-import firrtl.{AnnotationSeq, MemKind, PortKind}
+import firrtl.{AnnotationSeq, MemKind, PortKind, RegKind}
+import logger.LazyLogging
 import treadle._
 import treadle.chronometry.{Timer, UTC}
 import treadle.utils.Render
 import treadle.vcd.VCD
 
 import scala.collection.mutable
+import scala.util.Random
 
 //scalastyle:off magic.number number.of.methods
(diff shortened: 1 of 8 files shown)


chick:
@ekiwi Do you mind taking a look at this. It's my first attempt at supporting randomization

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -153,6 +157,43 @@ class ExecutionEngine(
     }
   }
 
+  def randomize(): Unit = {
+    val rand = new Random
+
+//    val symbolsToDo: Seq[Symbol] = {
```
ekiwi-sifive, line 163: you should remove this code or uncomment it

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -153,6 +157,43 @@ class ExecutionEngine(
...
+//      symbolTable.registerNames.map(name => symbolTable(name)).toSeq ++
+//        symbolTable.moduleMemoryToMemorySymbol.values.flatten
+//    }
+    val symbolsToDo: Seq[Symbol] = symbolTable.symbols.toSeq
+
+    symbolsToDo.foreach { symbol =>
+      rand.setSeed(symbol.name.hashCode + userRandomSeed)
```
ekiwi-sifive, line 170: Are you sure that reseeding the random number generator is a good idea in terms of getting "good" random results?
If you can assure that all values are generated in a deterministic order, e.g. by sorting all symbols by their name, they should be deterministic and only require the random number generated to be seeded a single time.

src/test/scala/treadle/RandomizeCircuit.scala
```diff
@@ -0,0 +1,147 @@
...
+  val t = new Timer()
+
+  def method1: Unit = {
+    Random.setSeed(System.currentTimeMillis())
+    t("method1") {
+      val a = Array.tabulate(1000) { i =>
+        Random.nextInt()
+      }
+      val sum = a.sum
+      println(s"method1 sum $sum")
```
ekiwi-sifive, line 37: There shouldn't be prints in a unit test.

src/test/scala/treadle/RandomizeCircuit.scala
```diff
@@ -0,0 +1,147 @@
...
+    method1
+    method2
+    method1
+    method2
+    t.clear()
+    for (i <- 0 to 10) {
+      method1
+      method2
+    }
+    println(t.report())
```
ekiwi-sifive, line 62: can you turn the check into assertions instead of printing to the command line?

src/test/scala/treadle/RandomizeCircuit.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    tester.step(10)
+
+    tester.reset(30)
+    tester.step(10)
+
+    tester.randomize()
+    tester.step(10)
+
+    tester.finish
```
ekiwi-sifive, line 143: Is there something you can assert here? Since the random numbers should be stable, asserting that they are different from zero or something like that should be ok and not make the tests flaky.

src/main/scala/treadle/executable/ExecutionEngine.scala
chick: It's a good question, I'll look into it

chick:
Thanks for the review @ekiwi-sifive I'll clean up all of these things. I should have been more specific, before I do much more work here I was hoping you could opine on whether this is a good approach or not, in particular I'm curious about
- Is it useful to be able to re-randomize in the current run.
- Is the registers and the memory locations everything that needs to be done. 

With respect to random seeds, I did the reseeding because i thought it would be useful to have the registers, for instance, always get randomized the same (for a given user seed) even if changes or additions were made to the registers in the circuit. Maybe that doesn't matter but I thought it might.  If it doesn't it shouldn't be hard to do a single set seed then generated the random values is a specific order

ekiwi-sifive:
>  I was hoping you could opine on whether this is a good approach or not

I see. I guess one of the things I would be interested in is: What are to use cases for `treadle` that do not involve `chisel-test`? I do believe that `chisel-test` should implement its own randomization (via register and (maybe) memory pokes) in order to be consistent accross simulator backends.

ekiwi-sifive:
Some for things that need to be randomized (probably not a complete list):
- result of a division (or `rem`) by zero
- Memories:
  - reading from a read port when `en` is false should return a random result
  - read-under-write (i.e. reading from the same address that you are writing to) on a Memory where `read-latency > 0` and the read-under-write behavior is undefined should return a random result on the read port
  - write conflicts (i.e. writing to the same address from more than one write port in a single clock tick) should randomize the destination
  - out-of-bounds read (can happen when the `depth` is not a power of two) should return a random result even when `en` is true

ekiwi-sifive:
>     * Is it useful to be able to re-randomize in the current run.

Might be useful for testers, but seems kind of hard to integrate for now since testers does not even cache the Verilator C++ simulator afaik.

>     * Is the registers and the memory locations everything that needs to be done.

See comment above. There are some things that might need to be randomized at runtime. Keep that in mind when designing the randomization algorithm, but no need to but all of that into this PR.

>  If it doesn't it shouldn't be hard to do a single set seed then generated the random values is a specific order

That is what I would try.

src/test/scala/treadle/RandomizeCircuit.scala
chick: Removed

src/test/scala/treadle/RandomizeCircuit.scala
chick: change to check values

src/main/scala/treadle/executable/ExecutionEngine.scala
chick: removed code

src/test/scala/treadle/RandomizeCircuit.scala
chick: changed to test of quality here.

ekiwi-sifive:
Do you know how this is going to interact with https://github.com/freechipsproject/treadle/pull/220 ?

chick:
I am not sure what it should do but currently if you employ randomization with the new Annotation it will randomize the circuit before any memories are loaded. If you somehow manage to call randomize later it will overwrite memories with random values. Do you think that is the correct behavior

ekiwi-sifive:
> if you employ randomization with the new Annotation it will randomize the circuit before any memories are loaded.

That is the correct behavior!

PR #218, opened 2020-07-02
fix build for mill

sequencer:
This PR fix the mill build which failed for a long time.
It can be work with freechipsproject/chisel3#1504
Request review from @chick


sequencer:
@chick request for review.

PR #220, opened 2020-07-14
Provides support for latest memory initialization annotations

chick:
MemoryArrayInitAnnotation, MemoryRandomInitAnnotation, MemoryScalarInitAnnotation
- Execution engine has new method referenceTargetToSymbols
  - Finds all symbols descrived by a reference target
- Add code to initialize memory based on these annotations in Memory.scala
- Add new test function to demonstrate initialization
- Allow for partial paths in reference
- Expanded tests to handle the partial paths

### Build Dependencies
Build with: git-clone firrtl https://github.com/freechipsproject/firrtl.git master

### DIFF
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -369,11 +369,14 @@ class ExecutionEngine(
   def referenceTargetToSymbols(referenceTarget: ReferenceTarget): Seq[Symbol] = {
     if (referenceTarget.path.nonEmpty) {
       // a specific path into the circuit
-      val name = referenceTarget.path.map { case (Instance(name), _) => name }.mkString(".") + "." + referenceTarget.ref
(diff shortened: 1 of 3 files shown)


ekiwi-sifive:
Do you think you could also add support for the `Preset` annotation (if you haven't done so yet)?
It essentially does the same thing as the memory init annotation, but for registers (and the init value is just the reset value).

https://github.com/freechipsproject/firrtl/blob/5c0c0018d812d57270035a9d3bd82e2289acf4ec/src/main/scala/firrtl/annotations/PresetAnnotations.scala

ekiwi-sifive:
`PresetAnnotationSpec`: https://github.com/freechipsproject/firrtl/blob/ba1f24345ac5ab20c669c73b871920001ac3a8ed/src/test/scala/firrtlTests/PresetSpec.scala

davidmlw:
I have just used this feature. Very cool!

PR #228, opened 2020-07-21
Remove Driver. Follow-up to #226

chick:
This PR does more cleanup work
- Gets rid of all options managers in Treadle
  - The TreadleOptionsManager
  - The Repl options manager
  - The vcd options manager
- Removes a couple of other references to deprecated things


ekiwi-sifive:
@chick can you rebase this on top of the latest master branch please?

chick:
@ekiwi-sifive I think I am going to wait to a bit before merging this because it will require upstream modifications to deal with the lack of a Driver/OptionsManager for treadle. So you can hold off on reviewing also

src/main/scala/treadle/vcd/VCD.scala
```diff
@@ -10,10 +10,18 @@ import logger.LazyLogging
...
 import scala.collection.mutable.ArrayBuffer
 import scala.util.matching.Regex
 
-object VCD extends LazyLogging {
+/**
+  * This effective stage main exercises vcd reading and optionally writing
+  * and depending up filtering options can pull out only those change values that
+  * are specific to a particular module
+  */
+object VCD extends StageMain(new VcdStage) with LazyLogging {
```
ekiwi, line 24: @chick Why does the VCD object have to extend `StageMain`?
I saw that you added a `VCDStage` already. What is the difference?

ekiwi: Could we move this into a separate object?
Like maybe just: `object VCDStageMain extends StageMain(new VcdStage) {}`?

ekiwi: Then also move it into `VcdStage.scala`.

chick: @ekiwi afaik it's just a pattern. `StageMain` provides a `main` that parses the CLI args and calls the associated stage's `execute` method. The pattern would ordinarily have a `VcdMain` but the existing API is to just to have the object`VCD` have the `main`. I guess I could create a `VcdMain` as the new canonical way of calling it. I'm not sure how I would deprecate the existing `VCD#main`. Do you think it's worth doing that.

ekiwi: Oh, I see. So this is done for backwards compatibility reasons. That sounds OK!

ekiwi (approved):
:shipit:

PR #234, opened 2020-08-04
Implement a finish call for engine

chick:
This calls ScalaBlackBox#finish method allowing
for cleanup or file writing to take place at end of simulation.


davidmlw (approved):
I have get these run in one of my project.

PR #235, opened 2020-08-04
enable Scala 2.13 cross compilation

ekiwi-sifive:


ekiwi-sifive:
@chick do I see it correctly that github actions will run tests for all enabled Scala versions by default?

chick:
This was done via PR #290

PR #238, opened 2020-08-10
Fix problems caused by PadWidths creating Dshlw

chick:
- Switch to using Forms.LowForm
- Include other optimizations from
  - LowFormMinimumOptimized
  - LowFormOptimized
  - without those passes some treadle tests take 25X longer to run


ekiwi-sifive:
Does this solve the problem? It seems like `PadWidths` would still be scheduled since some of the optimization passes depend on `PadWidths`.

ekiwi-sifive:
@chick  if you could add the test case where you see a Dshlw to the treadle test suite, I could have a look at it.

chick:
@ekiwi-sifive @jackkoenig could I get an review of this please. It's one of the pieces needed for dshlw fix, other pr's are approved.

PR #246, opened 2020-08-30
Allows chisel-iotesters to turn on vcd logging of memories

chick:
This adds two options that were added to newer stage phase CLI into the Driver legacy code.
Add deprecation information to treadle CLI for deprecated flags


ekiwi:
In the interest of keeping things simple: would it make sense to deprecate the "Driver legacy code" in `1.4` so that you can remove it before `1.5`? @chick

PR #250, opened 2020-09-23
Update .scalafmt.conf to be the same as firrtl

chick:
update all files in src/ and build.sbt and build.sc
remove scalastyle files


ekiwi (approved):
Seems like a good idea to me.

PR #262, opened 2020-10-27
Remove fossil junit dependency

chick:
Appears to have been a legacy hack for an earlier version of sbt
removed from build.sbt and build.sc


chick:
Too far behind

PR #263, opened 2020-11-03
Statement/Branch coverage for all tests

Dobios:
Adding Statement/Branch Coverage to Treadle  
---------------
The idea was to modify the treadle tester so that it would make certain modifications to the given FIRRTL source code in order to track the line coverage. Once that was done, we needed to check these additional ports and compile the results in a line coverage report. The modifications related to this work can be found [here](https://github.com/chisel-uvm/treadle/blob/master/src/main/scala/treadle/coverage/CoverageParser.scala). Note that coverage can be enabled using the [EnableCoverage](https://github.com/chisel-uvm/treadle/blob/master/src/main/scala/treadle/TreadleOptions.scala) annotation.
  
### Modifying the input FIRRTL code  
The first thing that needed to be done was to add ports to the given design. This is done depending on the amount of multiplexers that are in the design, since any code that isn't in a multiplexer is executed by default. The idea would thus be to do the following:  
- Parse the input FIRRTL code to find:  
  1. The location of the i/o declaration  
  2. The locations of every multiplexer in the design  
  3. The total number of multiplexers  
- Add additional _coverage validator_ outputs in the i/o declaration part of the code. The idea is to add 2 additional 1-bit ports per multiplexer.  
- At each multiplexer use in the FIRRTL source, output the path taken to 2 coverage validators as follows:  
  ```scala  
  out_a <= mux(cond, in_a, in_b)  //Example use of a multiplexer  
  io_cov_valid_1 <= cond  
  io_cov_valid_2 <= not(cond)  //Use of coverage validators to keep track of the path that was taken  
  ```    
    
These are all of the modifications that are needed to keep track of line coverage during a simulation.  
  
### Generating a Coverage report  
The main work here was interpreting the meaning of the outputs added to the FIRRTL source. The idea here was to do a second pass of the coverage parser once a given test was done. The big difference now is that we have access to the values of the different _coverage validators_, which allows us to know which multiplexer paths were taken during the different tests. So we can compile the data retrieved from the _coverage validators_ througout an entire test suite to find out which paths were taken and which were skipped. The results are then shown in a coverage report that has the following structure:  
```scala  
COVERAGE: 50.0% of multiplexer paths tested
COVERAGE REPORT:

+ circuit Test_1 :
+   module Test_1 :
+     input io_a : UInt<1>
+     input io_b_0 : UInt<2>
+     input io_b_1 : UInt<2>
+     input clock : Clock
+     output io_cov_valid_0 : UInt<1>
+     output io_cov_valid_1 : UInt<1>
+     output out : UInt<2>
+   
+     io_cov_valid_0 <= io_a
-     io_cov_valid_1 <= not(io_a)
+     out <= mux(io_a, io_b_0, io_b_1)  
```
Here the first line gives a general percentage of how many of the possible control paths were taken and then the `COVERAGE REPORT` shows a modified version of the FIRRTL source with coverage prefixes added to each line:  
- `+` means that the line was covered by a test in the test suite.  
- `-` means that the line was missed during all of the test suite's tests.  
  
#### Interpreting the Coverage Report  
What this report tells us (in the above example) is that one of our artificially added FIRRTL lines wasn't covered. This information can be used to deduce which path of our multiplexer was taken. For example, the above `COVERAGE REPORT`:  
```scala
+     io_cov_valid_0 <= io_a
-     io_cov_valid_1 <= not(io_a)
+     out <= mux(io_a, io_b_0, io_b_1)  
```  
actually tells us that the case where `io_a == 0` was never tested and thus one of our potential outputs was never tested.  
  
### Mapping loFIRRTL to Chisel  
One could discuss the usefulness of having line coverage on Chisel code. One main case where this would not be as helpful would be when writing hardware generators, since there the notion of coverage isn't as clear: _Has the hardware generator been tested?_ or more interestingly _Has the generated hardware been tested?_. I think that this question is open to discussion. This coverage report is thus given at the FIRRTL level but an other interesting report would contain the same line coverage information, shown directly in our original source description.  
   
***Future work*** on this part of the project could thus be to find a way to map the results to the original Chisel source using source locators. An other version of the coverage report would thus be:  
```scala
COVERAGE: 50.0% of multiplexer paths tested
COVERAGE REPORT:

+ circuit Test_1 :
+   module Test_1 :
+     input io_a : UInt<1>
+     input io_b_0 : UInt<2>
+     input io_b_1 : UInt<2>
+     input clock : Clock
+     output io_cov_valid_0 : UInt<1>
+     output io_cov_valid_1 : UInt<1>
+     output out : UInt<2>
+   
+     io_cov_valid_0 <= io_a
-     io_cov_valid_1 <= not(io_a)
+     out <= mux(io_a, io_b_0, io_b_1)  
```  
```scala
NEW REPORT:

+ class Test_1 extends Module {
+    val io = IO(new Bundle {
+        val a = Input(UInt(1.W))
+        val b = Input(Vec(2, UInt(2.W)))
+        val out = Output(UInt(2.W))  
+    })
+    when(io.a) {
+        out := io.b(0)
-    }.otherwise {
-        out := io.b(1)
+    }
+ }
```


ekiwi:
Thanks for the PR Andrew, this is great!

I really like the thought you put into how to communicate the coverage results back to the user.

Some notes on how this could be improved:
- Ideally you should try to write a firrtl pass over the firrtl compiler data structures instead of working on the string representation directly. This will make some of the current short comings much easier to solve.
- Currently you name the outputs without checking whether that name is already used in the module. The firrtl compiler provides a `Namespace` class to help you automatically resolve any potential name clashes.
- You seem to connect the coverage information to outputs of the module, however, how is this going to work for a design that has a module hierarchy? Here the firrtl compiler also features some helpers (namely the WiringTransform which can connect arbitrary wires/outputs throughout the module hierarchy).
- An interesting variation on your current approach for connecting the coverage signals might be to use the new `cover` statement. This way you would not have to worry about finding a unique name (cover statements are not named) or wiring things through the hierarchy. Unfortunately, treadle does not have support for cover statements yet, however, it is on the radar to implement and I think it would be great if you could give @chick some feedback on what you would expect from an implementation. For example, assuming that you have a pass to insert cover statements for all mux conditions, in which format would you expect treadle to report the results so that you can then use them to generate your coverage report?

For high firrtl / Chisel coverage it might be interesting to consider the pass that Daniel Kasza has been working on: https://github.com/danielkasza/dank-formal/blob/main/src/main/scala/transforms.scala#L13 He is essentially adding cover statements for all branches in high firrtl. This information should be enough to calculate the line and branch coverage and you could try to map it back to the Chisel source using the `FileInfo` that is attached to every firrtl statement.

Dobios:
Thanks a lot for the quick response!

I completely agree with the comments, this solution felt a bit brute-force-y when I implemented it.
I'll try to change my solution so that it uses the firrtl compiler rather than string parsing, as you suggested.Once that's done hopefully other problems will be easier to solve. 

Thanks again for the review!

ekiwi:
> Thanks a lot for the quick response!

Thanks for the PR!

> I completely agree with the comments, this solution felt a bit brute-force-y when I implemented it.
> I'll try to change my solution so that it uses the firrtl compiler rather than string parsing, as you suggested.Once that's done hopefully other problems will be easier to solve.

I know that writing your first firrtl pass is hard since there is very little documentation. Feel free to ask any questions that come up on the firrtl gitter: https://gitter.im/freechipsproject/firrtl

src/main/scala/treadle/Driver.scala
```diff
@@ -11,6 +11,7 @@ import scala.collection.mutable
 //scalastyle:off magic.number
 case class TreadleOptions(
   writeVCD:           Boolean = false,
+  enableCoverage:     Boolean = false,
```
chick, line 14: To avoid mid-release binary compatibility problems, I think you need to move enableCoverage to the end of the
the parameters for this class.

README.md
```diff
@@ -133,4 +133,4 @@ Keep in mind that there is no set of rules that will satisfy everyone.
 
 ### About ports and names
 The firrtl transformations that result in LoFirrtl alter the names of ports.
-What would be io.a becomes io_a and so forth.
+What would be io.a becomes io_a and so forth.
```
chick, line 136: I think is just white space at end of line.
You should run scalafmt on all the files you change.
But I just discovered that the scalafmt plugin is not enabled, will fix that.

chick (changes requested):
This looks great to me. Thanks for effort, it looks like a great contribution.
I will defer to Kevin on the technical analysis. I'd also like to maintain the scalafmt
formatting also.

README.md
chick: Actually, for the moment don't worry about scalafmt, it's about to undergo a version change and there's no reason to pollute your PR with formatting stuff.

Dobios:
New Version of the Statement Coverage PR
===================
As per the request, the coverage parser that was used to add additional coverage ports to the source code, via string parsing, was converted into a firrtl pass.

## FIRRTL pass  
A firrtl pass [AddCoverageExpressions](https://github.com/chisel-uvm/treadle/blob/master/src/main/scala/treadle/coverage/pass/AddCoverageExpressions.scala) was created, which takes care of adding the additional coverage ports that were previously done using brute-force string parsing.   
This now uses the `Namespace` of the module it's working on to make sure that no name collisions can occur and also handles multi-module circuits.  

## Coverage package object
The [coverage package object](https://github.com/chisel-uvm/treadle/blob/master/src/main/scala/treadle/coverage/package.scala) defines a new Info child class that allows us to store the names of the newly added ports in the circuit's info. These names are then used when compiling the final report (which is still done by modifying the original source code's string for printing purposes).   
The package also defines a `Ledger` class, similar to the one shown in the firrtl wiki. This class is used to keep track of the number of multiplexers in the design as well as maintaining references to the ports that were added in order to add them to the circuit's info at the end of the pass.  
  
## Coverage  
The [Coverage](https://github.com/chisel-uvm/treadle/blob/master/src/main/scala/treadle/coverage/Coverage.scala) class, which was previously named `CoverageReporter`, now uses the information stored in the circuit's info field to generate the final coverage report rather than parsing the source string a bunch of times.   
  
Let me know if ever there's anything that still needs to be changed/improved !

src/test/scala/treadle/CoverageTest.scala
```diff
@@ -0,0 +1,140 @@
...
+  it should "Get 100% coverage" in {
+    basicTest()
+  }
+
+  it should "Get 50% coverage" in {
+    halfCoverageTest()
+  }
+
+  it should "Get 100% coverage if no multiplexers are used" in {
+    noMuxCoverageTest()
```
ekiwi, line 138: Where do you assert that it actually returns 100% coverage?

src/main/scala/treadle/stage/phases/GetFirrtlAst.scala
```diff
@@ -29,8 +31,15 @@ object GetFirrtlAst extends Phase {
     def handleFirrtlSource(): Option[AnnotationSeq] = {
       annotationSeq.collectFirst { case FirrtlSourceAnnotation(firrtlText) => firrtlText } match {
         case Some(text) =>
-          val circuit = Parser.parse(text)
+
+          //Run the additional coverage firrtl passes if needed
+          val circuit = if (annotationSeq.contains(EnableCoverageAnnotation)) {
```
ekiwi, line 36: This should probably be done through a `RunFirrtlTransformAnnotation` instead of checking for your annotation manually. @chick probably has a more precise answer on how to integrate this with the existing treadle infrastructure.

src/main/scala/treadle/coverage/pass/AddCoverageExpressions.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import scala.collection.mutable
+
+/**
+  * Adds additional coverage statements to the Low FIRRTL source at every mux location.
+  * These statements can then be used to gather statement coverage information from inside of the Treadle tester
+  */
+object AddCoverageExpressions extends Pass {
+
+  override def prerequisites: Seq[TransformDependency] = firrtl.stage.Forms.LowForm
```
ekiwi, line 20: You need to specify your invalidates.

Probably something similar to this:
https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/RemoveAccesses.scala#L27

ekiwi: You also invalidate `InferTypes` since the reference you create in your `Ledger` does not carry a type.

src/main/scala/treadle/coverage/Coverage.scala
```diff
@@ -0,0 +1,154 @@
...
+            validIndex + 1
+          )
+        }
+      }
+    }
+
+    val firrtlSourceList = circuit.serialize.split("\n").toList
+
+    //Compute and print out the final coverage percentage
+    println("COVERAGE: " + getCoverage(tester.lineValidators) * 100 + "% of multiplexer paths tested")
```
ekiwi, line 141: You probably want to return the results as a `case class` and provide a separate serialization method in case users want to actually display the results.

This will allow you to have unit tests that do not print anything to the command line and you can automatically assert that the coverage was 100% without string parsing.

src/main/scala/treadle/coverage/pass/AddCoverageExpressions.scala
```diff
@@ -0,0 +1,87 @@
...
+  override def name = "Coverage!"
+
+  /**
+    * Run the coverage extension on every module
+    * @param c the circuit on with we want to add the coverage extensions
+    * @return the newly modified version of the circuit including coverage validators
+    */
+  override def run(c: Circuit): Circuit = {
+    val ledger = new Ledger()
+    val newModule =  c.modules.map(coverM(ledger))
```
ekiwi, line 31: You should probably check that the number of modules is one and throw an exception otherwise since the way that you are currently adding ports will only for for the top level module.
Modules that are lower in the hierarchy would need their coverage ports to somehow be wired to the top-level which is not implemented atm.

Dobios:
Thanks a lot for the review @ekiwi , your comments make a lot of sense, I'll get right to it!

src/main/scala/treadle/stage/phases/GetFirrtlAst.scala
chick: I think Kevin is right here and I think it's pretty simple to change.
If you do something like the following
Add this to AddCoverageExpressions
```
  override def optionalPrerequisiteOf: Seq[TransformDependency] = Seq(Dependency[BlackBoxSourceHelper])
```
Add 
```
    Dependency(AddCoverageExpressions),
```
to line 18 of PrepareAst, then revert the changes to `GetFirrtlAst` as the other changes will have scheduled your pass.

chick (changes requested):
This is looking pretty good. Check out my suggestions in comments for GetFirrtlAst

chick:
I missed one other suggestion for how to schedule and run the pass. I think you should add this to AddCoverageExpressions
```scala
  /** Only run this pass if there is an EnableCoverageAnnotation present
    *
    * @param state
    * @return
    */
  override def execute(state: CircuitState): CircuitState = {
    if(state.annotations.contains(EnableCoverageAnnotation)) {
      state.copy(circuit = run(state.circuit))
    } else {
      state
    }
  }
```

Dobios:
Thanks for the comments @chick ! Does this current state look good?

src/main/scala/treadle/stage/phases/GetFirrtlAst.scala
```diff
@@ -29,8 +29,11 @@ object GetFirrtlAst extends Phase {
     def handleFirrtlSource(): Option[AnnotationSeq] = {
       annotationSeq.collectFirst { case FirrtlSourceAnnotation(firrtlText) => firrtlText } match {
         case Some(text) =>
+
+          //Run the additional coverage firrtl passes if needed
```
chick, line 33: I think this comments got left behind after last round. I think GetFirrtlAst doesn't need any changes now that the pass is schedule using the standard stage/phase mode.

chick (changes requested):
One last quibble on GetFirrtlAst.scala.
Fix that and I will approve.
Otherwise looks great. 
@ekiwi any other thoughts

ekiwi:
>  @ekiwi any other thoughts

I think it is a fine implementation for the time being. Maybe we could mark it as experimental to allow for future API breaks.

The next step would require us to implement support for the cover statement in treadle. This way we can replace the output ports with cover statements which would almost automatically make this work for submodules as well. @Dobios work can serve as a good use-case example to see how a custom coverage report generator might want to plug into a generic cover statement implementation. I.e., the idea would be that treadle can load coverage report generator classes that analyze the coverage and turn it into a human readable report.

src/main/scala/treadle/stage/phases/GetFirrtlAst.scala
Dobios: woops yeah I didn't see that there, my bad ^^"

Dobios:
> > @ekiwi any other thoughts
> 
> I think it is a fine implementation for the time being. Maybe we could mark it as experimental to allow for future API breaks.  

Nice thanks!

> The next step would require us to implement support for the cover statement in treadle. This way we can replace the output ports with cover statements which would almost automatically make this work for submodules as well. @Dobios work can serve as a good use-case example to see how a custom coverage report generator might want to plug into a generic cover statement implementation. I.e., the idea would be that treadle can load coverage report generator classes that analyze the coverage and turn it into a human readable report.

I did see that since I've implemented the first version of this the cover statement was added to Chisel, it does seem like a more elegant solution. Is anyone currently working on adding it to Treadle?

chick (approved):

jackkoenig:
To backport this to 1.3.x, we should tweak the change to `TreadleOptions` to use https://github.com/alexarchambault/data-class

Basically, define the `copy` and `unapply` methods as they exist today (well, before this change), and then change it from a `case class` to `@data`.

ekiwi:
> To backport this to 1.3.x, we should tweak the change to `TreadleOptions` to use https://github.com/alexarchambault/data-class

I don't think this needs to be backported. This is more of a technology demo and not really ready for production usage anyways. Like currently it only works if you only have a single module in your design.

jackkoenig:
Makes sense, fair enough. Getting the SNAPSHOT autopublishing in will be nice so people can try it out using the SNAPSHOT.

PR #267, opened 2020-11-18
Simplify Github Actions CI

jackkoenig:
WIP


chick (approved):
That's a very nice simplification

jackkoenig:
A lot of the code before was dealing with using a Docker image for dependencies like Verilator; none of which are needed by this repo. I suspect we'll want to refer back to it if we were to move Chisel or FIRRTL to Github Actions.

That being said, we might as well make it as simple as possible for Treadle since we have that luxury 🙂

jackkoenig:
Also some of it was for building dependencies which becomes unnecessary with the auto SNAPSHOT publishing

jackkoenig:
@Mergifyio backport 1.3.x 1.2.x

PR #275, opened 2020-11-24
Update version to 1.5-SNAPSHOT, skip 1.4

jackkoenig:
This brings the major version of Treadle in line with Chisel and FIRRTL.


chick (approved):
lgtm

PR #278, opened 2020-12-03
Fix milestone in .mergify.yml

jackkoenig:
This was probably copied from FIRRTL, but the actual milestone is `1.2.x`, not `1.2.X`


chick (approved):
lgtm

PR #284, opened 2020-12-08
Fix settings so maven reports treadle as apache-2.0

chick:
- Fixes mill's build.sc to refer to apache license

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -1,12 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Build script for mill 0.6.0
-import mill._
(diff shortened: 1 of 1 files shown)


jackkoenig (approved):

build.sc
```diff
@@ -1,12 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Build script for mill 0.6.0
-import mill._
```
sequencer, line 4: These import should not be removed.

chick: I put them back, I don't know how I removed them in the first place.

PR #286, opened 2020-12-09
Fix mill and add GitHub CI for mill

sequencer:


sequencer:
@chick Need to test if CI works fine. I think I need to be grant permission to push this to `chipsalliance/treadle`, since my fork won't be tested by CI.

chick:
You should have write permission now

chick (approved):
lgtm

.github/workflows/test.yml
```diff
@@ -19,6 +19,11 @@ jobs:
         uses: olafurpg/setup-scala@v10
       - name: Cache
         uses: coursier/cache-action@v5
+      - name: Setup Mill
+        uses: jodersky/setup-mill@master
```
jackkoenig, line 23: It would be better to pin this to a version
```suggestion
      - name: Setup Mill
        uses: jodersky/setup-mill@v0.2.3
```

jackkoenig (changes requested):

PR #290, opened 2020-12-18
Add support for Scala 2.13

jackkoenig:
Treadle never needed `-Xsource:2.11`, so I deleted it. It also doesn't need macros-paradise, so I deleted that from the Mill build.

Ironically, this is _almost_ binary compatible for 2.11 and 2.12, but deleting the import in `src/main/scala/treadle/ScalaBlackBox.scala` changes some return types from `collection.Set` to `collection.immutable.Set`. This change could possibly be made in a fully binary incompatible way, but it's not really needed until 1.5 (since Chisel almost certainly will break binary compatibility in adding 2.13). I figured it was better to fix that API because I suspect that using `collection.Set` was unintentional unless @chick tells me otherwise.

### DIFF
--- a/src/main/scala/treadle/executable/SymbolTable.scala
+++ b/src/main/scala/treadle/executable/SymbolTable.scala
@@ -210,12 +210,7 @@ object SymbolTable extends LazyLogging {
     var printfCardinal: Int = 0
     val lastPrintfInMOdule = new mutable.HashMap[Module, Symbol]
 
-    val moduleMemoryToMemorySymbol = new mutable.HashMap[String, mutable.HashSet[Symbol]] {
-      override def default(key: String): mutable.HashSet[Symbol] = {
-        this(key) = new mutable.HashSet[Symbol]()
-        this(key)
-      }
-    }
+    val moduleMemoryToMemorySymbol = new mutable.HashMap[String, mutable.HashSet[Symbol]]()
 
     val blackBoxImplementations = new mutable.HashMap[Symbol, ScalaBlackBox]()
 
@@ -404,7 +399,8 @@ object SymbolTable extends LazyLogging {
           }
           val moduleMemory = module.name + "." + defMemory.name
 
-          moduleMemoryToMemorySymbol(moduleMemory) += memorySymbols.head
+          val values = moduleMemoryToMemorySymbol.getOrElseUpdate(moduleMemory, new mutable.HashSet)
+          values += memorySymbols.head
 
         case stop @ Stop(info, _, clockExpression, enableExpression) =>
           getClockSymbol(clockExpression) match {
--- a/src/main/scala/treadle/utils/AugmentPrintf.scala
+++ b/src/main/scala/treadle/utils/AugmentPrintf.scala
@@ -41,7 +41,8 @@ class AugmentPrintf extends Transform with DependencyAPIMigration {
         case s: Stop =>
           val newStmts = mutable.ArrayBuffer[Statement]()
           val newStop = s.mapExpr(insert(newStmts, namespace, s.info, s.clk))
-          Block(newStmts :+ newStop)
+          newStmts += newStop
+          Block(newStmts.toSeq)
         case p: Print =>
           val newStmts = mutable.ArrayBuffer[Statement]()
           val newName = namespace.newTemp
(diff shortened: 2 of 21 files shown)


jackkoenig:
Incredibly annoyingly, fixing those warnings that the docs make errors the suggested way breaks the tests. Great.

src/main/scala/treadle/executable/SymbolTable.scala
```diff
@@ -210,12 +210,7 @@ object SymbolTable extends LazyLogging {
     var printfCardinal: Int = 0
     val lastPrintfInMOdule = new mutable.HashMap[Module, Symbol]
 
-    val moduleMemoryToMemorySymbol = new mutable.HashMap[String, mutable.HashSet[Symbol]] {
-      override def default(key: String): mutable.HashSet[Symbol] = {
-        this(key) = new mutable.HashSet[Symbol]()
-        this(key)
-      }
-    }
+    val moduleMemoryToMemorySymbol = new mutable.HashMap[String, mutable.HashSet[Symbol]]()
```
jackkoenig, line 213: The deprecation warning (an error when building the docs) says that `withDefaultValue` is the normal way to fix this, but it does **not** mesh with mutation. `withDefaultValue` never actually adds the key to the underlying `mutable.HashMap` so it's just sort of a mess, don't use it with mutable stuff.

chick: I think I've used this idiom in quite a few places. Too bad `withDefaultValue` doesn't work in the obvious way

src/main/scala/treadle/utils/AugmentPrintf.scala
```diff
@@ -41,7 +41,8 @@ class AugmentPrintf extends Transform with DependencyAPIMigration {
         case s: Stop =>
           val newStmts = mutable.ArrayBuffer[Statement]()
           val newStop = s.mapExpr(insert(newStmts, namespace, s.info, s.clk))
-          Block(newStmts :+ newStop)
+          newStmts += newStop
```
chick, line 44: What's the change being addressed by `Block(newStmts :+ newStop)` going to
 ```
newStmts += newPrint
 Block(newStmts.toSeq)
```
And in a previous case you checked the newStmts length and avoided a block. is there a reason for the different pattern here

**Review by chick — APPROVED** [review]
This all looks good. Seems like 2.13 catches a lot of little stuff that was not warned of before

File: src/main/scala/treadle/executable/SymbolTable.scala (hunk 1 above)
Comment (jackkoenig) [reply src/main/scala/treadle/executable/SymbolTable.scala:213]: Yeah it is a powerful tool to lose. One option is to have a custom class extend `trait mutable.Map` that is just backed up by a `mutable.HashMap`. There are 4 methods that we'd have to implement, but it's an option for cases where just switching to `getOrElseUpdate` is too onerous. In this case it was only a few places so I thought it was fine.

File: src/main/scala/treadle/utils/AugmentPrintf.scala (hunk 2 above)
Comment (jackkoenig) [reply src/main/scala/treadle/utils/AugmentPrintf.scala:44]: In Scala 2.12, `Seq` is `collection.Seq` (super type of all immutable *and* mutable Seqs). In Scala 2.13, `Seq` is `collection.immutable.Seq`. Thus we cannot pass `mutable.ArrayBuffer` where a `Seq` is required anymore. My change is functionality the same as what was there although I will grant it's a slightly different way to accomplish it. I could've just done:
```scala
Block((newStmts :+ newStop).toSeq)
```
That would've been 2 copies, the `:+` is a copy, followed by conversion to an immutable `Seq`, another copy. Since `newStmts` does not escape this scope, I figured we might as well just mutably append and have the 1 copy. Same number of copies without jumping through any hoops!

An alternative with no copies is to use `mutable.ListBuffer` which has a free `.toList` (it copies if you then mutate the underlying buffer again). The only issue is that `Lists` while decent, aren't great datastructures for memory use. Another alternative is to start using things like `VectorBuilder` in cases like this. I wanted to keep this PR simple so I avoided larger changes like that. My proposal is that we get everything on 2.13 and then switch it to be our default and start measuring performance and improving it.

jackkoenig: The earlier length check was already there I think, I did the same conversion.

PR #295, opened 2021-01-20
Fix Mergify condition for labeling backports

jackkoenig:
See https://github.com/chipsalliance/firrtl/pull/2048


chick (approved):
looks ok to me

PR #296, opened 2021-01-20
Proof of concept for coverage statement counting in Treadle

chick:
This creates a new treadle execution element, `VerifyOp`, that keeps count of all the number of times a coverage statements has seen a rising edge of its clock and how many times the enable and predicate are both true.

At the end of the run the a CSV text file is created in the working directory <circuit-name>.coverage.txt

Additionally a small pretty printer, `CoveragePrettyPrinterMain`, that can take the coverage CSV file and the original firrtl file and produces a text rendering of the firrtl with coverage info added at the end of the lines of coverage statements.
By default those lines are color code as follows
- red means no coverage hits
- cyan means hits but not always
- green means coverage hits equals

Specifically
- Add a couple of hooks to execution engine to help with coverage file generation
- Execution compiler adds VerifyOp's to the the execution stack
- Symbol Table builds lists of Coverage statements encountered
- ExpressionViewBuilder has handling code also
- Added some hints to disambiguate who can't handle a class of Firrtl statements

### DIFF
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -493,13 +493,13 @@ class ExecutionEngine(
   Currently this telling black boxes to finish up things if they need to
   and have the coverage counts recorded if they exist.
    */
-  def finish(): Unit = {
+  def finish(writeCoverageReport: Boolean = false): Unit = {
     symbols.foreach { symbol =>
       symbolTable.getBlackboxImplementation(symbol).foreach { blackBox =>
         blackBox.finish()
       }
     }
-    if (symbolTable.verifyOps.nonEmpty) {
+    if (writeCoverageReport && symbolTable.verifyOps.nonEmpty) {
       val text = symbolTable.verifyOps.map { verifyOp =>
         s"${verifyOp.info.toString.trim},${verifyOp.message.escape},${verifyOp.clockCount},${verifyOp.coverCount}"
       }.sorted.mkString("\n")
--- a/src/main/scala/treadle/executable/ExecutionEngine.scala
+++ b/src/main/scala/treadle/executable/ExecutionEngine.scala
@@ -2,13 +2,14 @@
 
 package treadle.executable
 
-import java.io.PrintWriter
+import java.io.{File, PrintWriter}
 
 import firrtl.annotations.ReferenceTarget
 import firrtl.annotations.TargetToken.Instance
 import firrtl.ir.{Circuit, NoInfo}
 import firrtl.options.StageOptions
 import firrtl.options.Viewer.view
+import firrtl.stage.OutputFileAnnotation
 import firrtl.{AnnotationSeq, MemKind, PortKind, RegKind}
 import logger.LazyLogging
 import treadle._
@@ -33,6 +34,11 @@ class ExecutionEngine(
 
   var vcdOption:   Option[VCD] = None
   var vcdFileName: String = ""
+
+  // use the generated file root to create files in working dir with desired suffixes
+  private val topName: String = annotationSeq.collectFirst { case OutputFileAnnotation(ofn) => ofn }.getOrElse(ast.main)
+  private val stageOptions = view[StageOptions](annotationSeq)
+  val generatedFileRoot: String = stageOptions.getBuildFileName(topName)
 
   val expressionViewRenderer = new ExpressionViewRenderer(
     dataStore,
@@ -79,8 +85,7 @@ class ExecutionEngine(
     scheduler.setVerboseAssign(verbose)
   }
 
-  /**
-    * turns on evaluator debugging.  Can make output quite
+  /** turns on evaluator debugging.  Can make output quite
     * verbose.
     *
     * @param isVerbose  The desired verbose setting
@@ -256,8 +261,7 @@ class ExecutionEngine(
     }
   }
 
-  /**
-    * Update the dataStore with the supplied information.
+  /** Update the dataStore with the supplied information.
     * IMPORTANT: This should never be used internally.
     *
     * @param name  name of value to set
@@ -328,8 +332,7 @@ class ExecutionEngine(
     value
   }
 
-  /**
-    * Update the dataStore with the supplied information.
+  /** Update the dataStore with the supplied information.
     * IMPORTANT: This should never be used internally.
     *
     * @param symbol symbol to set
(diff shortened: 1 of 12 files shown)


src/main/scala/treadle/executable/SymbolTable.scala
```diff
@@ -443,24 +453,51 @@ object SymbolTable extends LazyLogging {
...
+              addSymbol(verifySymbol)
+
+              verifyCardinal += 1
+              verifyToVerifyInfo(verify) = VerifyInfo(verifySymbol, verifyCardinal)
+              addDependency(verifySymbol, expressionToReferences(clockExpression))
+              addDependency(verifySymbol, expressionToReferences(enableExpression))
+              addDependency(verifySymbol, expressionToReferences(predicateExpression))
+
+              lastVerifyInModule.get(module) match {
+                case Some(lastPrintfSymbol) =>
```
ekiwi, line 483: I assume the `lastPrintfSymbol` is a copy and paste error.

Could you also elaborate on why you need to add a dependency on the previous cover statement?

chick: Yes, its a CAP typo. That code probably isn't necessary, it was derived from printf's very specifically caring about the order of evaluation. I guess that wouldn't apply here

chick: Thinking about it more, I can't think of why the order of coverage statement evaluation should matter. Can you. If not I'll just remove that code

ekiwi: It should not matter. For asserts it would though since two of them might be violated in the same cycle.

src/main/scala/treadle/executable/ExecutionEngine.scala
```diff
@@ -478,21 +480,33 @@ class ExecutionEngine(
...
         blackBox.finish()
       }
     }
+    if (symbolTable.verifyOps.nonEmpty) {
+      val text = symbolTable.verifyOps.map { verifyOp =>
+        s"${verifyOp.info.toString.trim},${verifyOp.message.escape},${verifyOp.clockCount},${verifyOp.coverCount}"
+      }.sorted.mkString("\n")
+      val writer = new PrintWriter(new File(generatedFileRoot + ".coverage.txt"))
+      writer.write(text)
+      writer.close()
```
ekiwi, line 508: Instead of writing this to a file, could you just collect the coverage into an annotation?

Something like:
`case class CoverageResultAnnotation(counts: Map[String, Int]) extends NoTargetAnnotation`

src/main/scala/treadle/executable/SymbolTable.scala
```diff
@@ -48,8 +48,10 @@ class SymbolTable(val nameToSymbol: mutable.HashMap[String, Symbol]) {
 
   val moduleMemoryToMemorySymbol: mutable.HashMap[String, mutable.HashSet[Symbol]] = new mutable.HashMap
 
-  val stopToStopInfo:   mutable.HashMap[Stop, StopInfo] = new mutable.HashMap[Stop, StopInfo]
-  val printToPrintInfo: mutable.HashMap[Print, PrintInfo] = new mutable.HashMap[Print, PrintInfo]
+  val stopToStopInfo:     mutable.HashMap[Stop, StopInfo] = new mutable.HashMap[Stop, StopInfo]
+  val printToPrintInfo:   mutable.HashMap[Print, PrintInfo] = new mutable.HashMap[Print, PrintInfo]
+  val verifyToVerifyInfo: mutable.HashMap[Verification, VerifyInfo] = new mutable.HashMap[Verification, VerifyInfo]
+  val verifyOps:          mutable.ListBuffer[VerifyOp] = new mutable.ListBuffer[VerifyOp]
```
ekiwi, line 54: Are these tables used for `cover` only or also for `assert`?

src/main/scala/treadle/executable/VerifyOp.scala
```diff
@@ -0,0 +1,51 @@
...
+  predicate:       IntExpressionResult,
+  enable:          IntExpressionResult)
+    extends Assigner {
+
+  var clockCount: Long = 0
+  var coverCount: Long = 0
+
+  def run: FuncUnit = {
+    val conditionValue = predicate.apply() > 0
+    val enableValue = enable.apply() > 0
```
ekiwi, line 31: Is there a reason for computing these unconditionally instead of only after checking the clock transition?

src/main/scala/treadle/executable/VerifyOp.scala
```diff
@@ -0,0 +1,51 @@
...
+
+    if (clockTransition.isPosEdge) {
+      clockCount += 1
```
ekiwi, line 34: I don't think the clock count is necessary and it is somewhat redundant since we will have many coverage statements that use the same clock.

src/main/scala/treadle/executable/ExecutionEngine.scala
ekiwi: Sorry, I meant `Map[String, Long]`, not `Int`.

chick: Sure, that would be easy. In general though simulators do not return annotations. I think we need to discuss how this should be implemented

src/main/scala/treadle/executable/SymbolTable.scala
chick: Only cover at present. It would be easy enough to extend. Would it be useful to record the first time, i.e. the clock transition count, that any of these statements is first true and first time its false.

src/main/scala/treadle/executable/VerifyOp.scala
chick: Great catch, I'll fix that

src/main/scala/treadle/executable/VerifyOp.scala
chick: Would it be useful to record the first time, i.e. the clock transition count, that any of these statements is first true and first time its false.

src/main/scala/treadle/executable/ExecutionEngine.scala
ekiwi: I looked at the `TreadleBackend` in `chiseltest` and I think it should be enough to add a `getCoverage: Map[String,Long]` method to the `TreadleTester` class.

src/main/scala/treadle/executable/VerifyOp.scala
chick: Deferred computing condition and enable values

src/main/scala/treadle/executable/SymbolTable.scala
chick: Fixed

ekiwi:
@chick I added the `getCoverage()` method to `TreadleTester`.

ekiwi (approved):
Once tests pass and Chick gives his final OK, we should :ship:

PR #298, opened 2021-02-17
Use Statement Names from Firrtl

ekiwi:
This PR makes `treadle` use the (optional) names of `printf`, `stop` and cover statements as their symbol name instead of auto-generating a name.
Statements that do not have a name (since the name is optional) are automatically named through the `EnsureNamedStatements` pass from firrtl.


chick (approved):
This looks good to me. I don't know of places where names are used much so I don't think there should be any compatibility problems with this PR

PR #317, opened 2021-03-26
Fix chisel3 import in README example

learning-chip:
`import chisel._` led to `not found: object chisel` error


edwardcwang (approved):

chick (approved):
LGTM, thanks for the PR

Issue #232, opened 2020-08-03
expectMemory should mention the index in the error Message


nrother:
The current message for a failed `expectMemory` is something like

> Error:expect(mem_a, 15) got 17

Source: https://github.com/freechipsproject/treadle/blob/master/src/main/scala/treadle/TreadleTester.scala#L376

It would be nice to not only mention the name of the memory ("mem_a"), but also the index that was read. For me, the current error message actually looks like 15 is the index, although it is the expected value (with 17 being the actual value).

chick:
Good idea. On my todo list. Thanks

chick:
@nrother hopefully addressed by #326

chick:
I believe this to be resolved

PR #326, opened 2021-05-10
expectMemory now shows index when error occurs

chick:
Helpful to see what memory index was being called when a failure occurs
Addresses Issue #232


ekiwi (approved):
:ship:

PR #331, opened 2021-05-28
Register presets

chick:
Evaluation support is a bit primitive.
When registers with async reset types are
matched with a RegisterPresetAnnoation then the
raw datastore is initialized with the literal
reset value.

### DIFF
--- a/src/main/scala/treadle/executable/ExpressionCompiler.scala
+++ b/src/main/scala/treadle/executable/ExpressionCompiler.scala
@@ -4,6 +4,7 @@ package treadle.executable
 
 import firrtl.PrimOps._
 import firrtl._
+import firrtl.annotations.ReferenceTarget
 import firrtl.ir._
 import treadle._
 import treadle.blackboxes.PlusArg
@@ -18,7 +19,8 @@ class ExpressionCompiler(
   validIfIsRandom:      Boolean,
   prefixPrintfWithTime: Boolean,
   blackBoxFactories:    Seq[ScalaBlackBoxFactory],
-  plusArgs:             Seq[PlusArg])
+  plusArgs:             Seq[PlusArg],
+  presetTargets:        Seq[ReferenceTarget])
     extends logger.LazyLogging {
 
   case class ExternalInputParams(instance: ScalaBlackBox, portName: String)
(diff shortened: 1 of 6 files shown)


src/main/scala/treadle/executable/ExpressionCompiler.scala
```diff
@@ -32,6 +34,19 @@ class ExpressionCompiler(
     }
   }
 
+  // Preset registers have references to the async reset
+  private val presetTargetList = presetTargets.map { presetTarget =>
```
ekiwi, line 38: We should add something like this to `firrtl`. I feel like I have written similar code before.
@seldridge @jackkoenig , do you know if there is already a way to turn a reference target into a path string?

src/main/scala/treadle/executable/ExpressionCompiler.scala
```diff
@@ -32,6 +34,19 @@ class ExpressionCompiler(
...
+    while (element.nonEmpty) {
+      val step = element.head._1.value
+      pathSteps += step
+      element = element.tail
+    }
+    pathSteps += presetTarget.ref
+    pathSteps.mkString(".")
+  }
+
   def getWidth(expression: Expression): Int = {
```
ekiwi, line 50: FYI, I think there is already a `firrtl.bitWidth` function that may do the same thing.

src/main/scala/treadle/executable/Symbol.scala
```diff
@@ -81,6 +82,11 @@ case class Symbol(
     f"$name%-40.40s $dataSize%3.3s $dataType%4.4s $bitWidth%6d " +
       f"$slots%6d $index%6d$dataSizeCode $cardinalNumber%6d $info"
   }
+
+  def matches(referenceTarget: ReferenceTarget): Boolean = {
+    val refString = referenceTarget.ref
```
ekiwi, line 87: what is the purpose of `refString`?

ekiwi (changes requested):
Instead of using the `PresetAnnotation` it would probably be much easier to use the `PresetRegAnnotation` annotation. The way this works is that the `PropagatePresetAnnotations` pass analyzes all reset signals and `PresetAnnotation`s and then it annotates all registers _directly_ with a `PresetRegAnnotation`. Maybe we will have to make that public for treadle to be able to access it. But I believe this will vastly simply your implementation since you are essentially re-implementing the `PropagatePresetAnnotations` apss.

ekiwi:
This is now blocked on: https://github.com/chipsalliance/firrtl/pull/2254

chick:
Closing this. Similar code was inadvertently pushed direct to master. By time it was discovered other changes had gone through. This will be cleaned up in next round of changes related to "Preset" changes

Issue #258, opened 2020-10-15
Figure out and fix MultiTopLevelClockSpec


chick:
Figure out and fix MultiTopLevelClockSpec

PR #337, opened 2021-06-11
Make MultiTopLevelClockSpec work

chick:
- Need better prediction system for when registers
- would transition
- Added tests of various clock relationships
Addresses Issue #258


ekiwi (approved):
LGTM

PR #342, opened 2021-07-29
Added coverage count test

BRF321JOJO:
This new test on coverage counts fails.
Coverage counts are expected to increase over time. Instead, they stay at 0 for all cycles.

This test passes when using VerilatorSimulator.

### DIFF
--- a/src/test/resources/ToggleCoverage.fir
+++ b/src/test/resources/ToggleCoverage.fir
@@ -1,0 +1,21 @@
+circuit test :
+  module test :
+    input metaReset : UInt<1>
+    input clock : Clock
+    input reset : UInt<1>
+    input cond : UInt<1>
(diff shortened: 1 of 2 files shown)


ekiwi:
@chick : This seems to be a bug in the evaluation order of cover counts and registers. Essentially it looks like the count looks at the new register value but it should actually look at the old register value when making the decision to increment.

ekiwi:
@BRF321JOJO Can you please run scalafmt on this PR? In `sbt` you need to execute `scalafmtAll`.

BRF321JOJO:
Amended the commit to include the proper formatting from scalafmtAll

ekiwi:
Now part of https://github.com/chipsalliance/treadle/pull/343

PR #343, opened 2021-08-02
Fix verification register handling

chick:
Fix coverage tests view of registers
- Like printf, verification (cover) statements need to use previous values of enable and predicate and arguments
- Add verification statement to AugmentPrintf transform
- For debugging show current value of register (still notes when register was not changed because of clock transition)


ekiwi (approved):
We talked about this offline and this is a good way to fix the issue for now! :ship:

chick:
This includes code from PR #342 and addresses the issue identified there (Thanks @BRF321JOJO).

PR #345, opened 2021-08-04
Fix preset related tests that snuck through

chick:
- Not sure how these got here but this
  - adds ignores to a couple in RegisterSpec
  - specifically looks for expected dependencies in SymbolTableSpec


ekiwi (approved):
LGTM

Issue #347, opened 2021-08-11
Include assertion failures in the VCD output


ekiwi:
Currently there is a signal for each assert in the VCD, however it remains low, even when the assertion fails.

In this example the firrtl is:
```
circuit test:
  module child:
    input clock : Clock
    input reset : AsyncReset
    output count : UInt<32>
    reg count_reg : UInt<32>, clock with : (reset => (reset, UInt(0)))
    count_reg <= add(count_reg, UInt(1))
    count <= count_reg

  module test:
    input clock : Clock
    input reset : AsyncReset

    inst c of child
    c.clock <= clock
    c.reset <= reset

    assert(clock, lt(c.count, UInt(5)), UInt(1), "") : leq_assert
```

When I replay the failure I get:
![Screenshot from 2021-08-11 13-14-05](https://user-images.githubusercontent.com/2456545/129097106-735e1ec4-dee0-4467-a357-713dbac44638.png)

I would like the `leq_assert` signal to go high in the cycle in which the assertion fails. Similar to:

![Screenshot from 2021-08-11 13-14-11](https://user-images.githubusercontent.com/2456545/129097241-4ff245c4-facc-4ea7-ba73-e87204aaad39.png)

PR #348, opened 2021-08-11
Draft PR for recording verification ops in VCD

chick:


chick:
@ekiwi Updated the tests so they actually look at the VCD output.
Just for fun, I implemented the cover statement to add the cover counts into the VCD file, but I don't
have a conviction that it's the right way to go

chick:
Addresses Issue #347

src/test/scala/treadle/VerificationSpec.scala
```diff
@@ -125,5 +127,110 @@ class VerificationSpec extends AnyFreeSpec with Matchers {
       }
       output.toString should include("input was not less that 0x7f")
     }
+
+    "asserts from verification should go high when triggered and be visible" in {
+      val targetDir = s"test_run_dir/verify_assumes_should_log_to_vcd"
```
ekiwi, line 132: this should be `asserts` not `assumes`

src/test/scala/treadle/VerificationSpec.scala
```diff
@@ -125,5 +127,110 @@ class VerificationSpec extends AnyFreeSpec with Matchers {
...
+          |
+          |  module test:
+          |    input clock : Clock
+          |    input reset : AsyncReset
+          |
+          |    inst c of child
+          |    c.clock <= clock
+          |    c.reset <= reset
+          |
+          |    cover(clock, lt(c.count, UInt(5)), UInt(1), "") : leq_assert
```
ekiwi, line 205: Maybe name this `leq_cover` instead of `leq_assert`

ekiwi (approved):
I think this is good for now. I would prefer for the assert signal to go high one cycle earlier, but we can try to address that in a followup PR. This PR already improves upon the status-quo.

Please address the two small comments and then feel free to merge.

PR #349, opened 2021-08-12
add suport for MemoryFileInlineAnnotation

ekiwi:


src/test/resources/ReadMem.hex
```diff
@@ -0,0 +1,3 @@
+// comment
+ab
+CD eF // any whitespace should be treated as a separator
```
chick, line 3: Did you mean for this to not have final newline as part of the test?

chick (approved):
Looks good to me.
One missing newline in hex data file, maybe that is intentional

src/test/resources/ReadMem.hex
ekiwi: Yes, this is intentional

PR #351, opened 2021-08-20
Treadle can find clock with some other name

chick:
This adds an attempt for treadle to find a top level clock with a name other than "clock"
by searching for one top level input with name clock.
This generally should only effect direct use of `TreadleTester` as most testing frameworks manage this on their own.
A test to demonstrate this change working

### DIFF
--- a/src/main/scala/treadle/TreadleTester.scala
+++ b/src/main/scala/treadle/TreadleTester.scala
@@ -4,8 +4,8 @@ package treadle
 
 import java.io.PrintWriter
 import java.util.Calendar
-
 import firrtl.AnnotationSeq
+import firrtl.ir.ClockType
 import firrtl.options.StageOptions
 import firrtl.options.Viewer.view
 import firrtl.stage.OutputFileAnnotation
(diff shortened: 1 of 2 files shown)


src/main/scala/treadle/TreadleTester.scala
```diff
@@ -74,8 +74,23 @@ class TreadleTester(annotationSeq: AnnotationSeq) {
...
+        symbol
+    }.toList
+  }
+
   val clockInfoList: Seq[ClockInfo] = if (clockInfo.isEmpty) {
-    if (engine.symbolTable.contains("clock")) {
+    val topClocks = findTopLevelClocks()
+
+    if (topClocks.length > 2) {
+      println(s"Warning: multiple top level clocks found without any ClockInfo information, is this intentional?")
```
ekiwi, line 88: Are you not using the logging infrastructure from `firrtl`?
Ideally you would use `logger.warn(...)` instead of `println(...)`

src/main/scala/treadle/TreadleTester.scala
```diff
@@ -74,8 +74,23 @@ class TreadleTester(annotationSeq: AnnotationSeq) {
 
   val startTime: Long = System.nanoTime()
 
+  private def findTopLevelClocks() = {
```
ekiwi, line 77: I guess this was the simplest place to implement this feature, but it does lead to multiple slightly different mechanisms of dealing with clocks. As an alternative, you could add `FindClocks` pass to the `TreadlePhase` which will automatically generate the appropriate clock annotations. That way everyone uses clock annotations instead of having this alternative way inside the tester.
Not a blocker for me, but maybe something to consider.

ekiwi (approved):

src/main/scala/treadle/TreadleTester.scala
chick: Created issue #353 to deal with this

src/main/scala/treadle/TreadleTester.scala
chick: Leaving as is for now.

PR #360, opened 2021-09-17
stop: report all stop statements that fail within a cycle

ekiwi:


ekiwi:
I am happy to update iotesters and chiseltest in case this breaks anything

chick (approved):
Looks good

PR #364, opened 2021-09-24
test: add test with a submodule to stop behavior spec

ekiwi:
This test would fail before https://github.com/chipsalliance/firrtl/pull/2367 was merged.


chick (approved):
Looks good to me

PR #366, opened 2021-09-29
vcd: exclude all signals that start with `_` by default

ekiwi:
This matches the default verilator behavior.


chick (approved):
This looks fine and verilator agreement is good. Just wondering if you think anyone expects the signals now ignored, I'd guess probably not

ekiwi:
This is based on user feedback, where a user was talking about how much nice Verilator generated VCDs are because they do not include the signals starting with `_`. Not sure if this will have unintended consequences. If users request the feature to be able to include `_` signals, we can always make an annotation for `chiseltest` which would translate to the correct `treadle` annotation or the correct Verilator flag depending on which backend is used.

PR #367, opened 2021-09-29
Fixes stops hiding other stops

chick:
Fixes stops hiding other stops
The communication between symbol table assembly and expression compiling used a case class as key
Changed the key to be a string representing the full pathname of the stop.
This seems to fix the problem.


ekiwi (approved):
Looks awesome! Thanks chick!

PR #368, opened 2021-09-29
Removed verbose mode, and vcd output from test

chick:
Missed commits from stop-fix
Removed verbose mode, and vcd output from test
Cleaned up imports


ekiwi (approved):
Thanks!

PR #382, opened 2021-12-13
Bump to sbt 1.5.6

jackkoenig:
Avoids any possibility of [CVE-2021-44228](https://github.com/advisories/GHSA-jfh8-c2jp-5v3q), although given this is Zinc's own logging, it would be pretty insane to see how that attack vector would manifest here.


chick (approved):
lgtm

PR #389, opened 2022-01-03
Downgreade ScalaTest to 3.1

schoeberl:


ekiwi:
I think the underlying issue was solved, and this PR is no longer needed.

PR #394, opened 2022-01-18
DataStore: ignore out of bounds writes to memory

ekiwi:
This seems to be a copy-paste error. The `Long` and `BigInt` versions already correctly implemented the wrapping.


ekiwi:
After investigating this a little bit more, it seems like the wrapping behavior is actually hiding a bug in a design that I am testing. It might be better to emulate Verilator behavior here and to just ignore out of bounds writes.

chick (approved):
LGTM

